PR #1, opened 2019-10-04
Create initial Docker setup

frwickst:


Dockerfile
```diff
@@ -0,0 +1,32 @@
+# ===============================================
+FROM helsinkitest/node:10-slim as appbase
```
Nizar-Rahme: Some of the packages I'm using require `node v12`, would it be possible to use it here?

midori0507: Apparently Node v12 is more stable than 11 so i would suggest node v12 also. 
Will require some work to put node 12 img on Hki dockerhub

Zeukkari:
Closing in favour of https://github.com/City-of-Helsinki/berth-reservations-admin/pull/7.

PR #2, opened 2019-10-14
VEN-181 | Set up dev dependencies

Nizar-Rahme:


.eslintignore
```diff
@@ -0,0 +1 @@
+src/serviceWorker.ts
```
midori0507: Do you want to snatch gitignore config from youth-ui or kukku ? Or in separate PR?

src/locales/en.json
```diff
@@ -0,0 +1,5 @@
+{
+  "harbours_list": {
```
midori0507: should we camelCase ?

.eslintignore
Nizar-Rahme: I guess it's the same that CRA generates.

src/locales/en.json
Nizar-Rahme: I don't know, any conventions we agreed on following?

.eslintignore
midori0507: Sorry, wrong file, thought this is gitignore

src/locales/en.json
midori0507: Not sure if we have convention on this, been camelCase in translation all my life :D Just asking

PR #3, opened 2019-10-15
VEN-236 | Fetch resources

Nizar-Rahme:


apollo.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  client: {
+    service: {
+      name: process.env.REACT_APP_API_URI,
+      url: 'https://venepaikka-api.test.hel.ninja/graphql_v2/',
```
midori0507: Probably not good to expose

PR #4, opened 2019-10-23
VEN-226 | Table component

Nizar-Rahme:


src/domain/harbors/HarborsPageContainer.tsx
```diff
@@ -0,0 +1,127 @@
...
+    },
+    {
+      Cell: ({ cell }: { cell: { value: number } }) => (
+        <PropertyIcon name="waterTap" disabled={!!cell.value} />
+      ),
+      Header: () => <PropertyIcon name="waterTap" />,
+      accessor: 'water',
+    },
+    {
+      Cell: ({ cell }: { cell: { value: number } }) => (
```
midori0507: Those thing are repeated, can you export that to something shorter ? function ? array.map rendering?

src/domain/harbors/HarborsPageContainer.tsx
```diff
@@ -0,0 +1,127 @@
...
+  if (loading) return <p>Loading...</p>;
+  if (error) return <p>Error</p>;
+
+  const tableData =
+    data && data.harbors
+      ? data.harbors.edges.reduce<object[]>((acc, harbor) => {
+          if (harbor && harbor.node && harbor.node.properties) {
+            const properties = harbor.node.properties.piers.edges.reduce<{
+              electricity: number;
+              gate: number;
```
midori0507: Extract typing separate out of this.

src/domain/harbors/HarborsPageContainer.tsx
```diff
@@ -0,0 +1,127 @@
...
+              lighting: number;
+              wasteCollection: number;
+              water: number;
+            }>(
+              (prev, pier) => {
+                if (pier && pier.node && pier.node.properties) {
```
midori0507: lodash get?

src/domain/harbors/HarborsPageContainer.tsx
```diff
@@ -0,0 +1,127 @@
...
+                  return {
+                    electricity: pier.node.properties.electricity
+                      ? prev.electricity + 1
+                      : prev.electricity,
+                    gate: pier.node.properties.gate ? prev.gate + 1 : prev.gate,
+                    lighting: pier.node.properties.lighting
+                      ? prev.lighting + 1
+                      : prev.lighting,
+                    wasteCollection: pier.node.properties.wasteCollection
```
midori0507: I suggest this data normalization should be in function, and with test / document. I cant even understand what's going on here with 2 level reduce like this.

src/domain/harbors/HarborsPageContainer.tsx
Zeukkari: I'm a bit torn on this. Lodash would improve readability of this, but I would prefer not adding an extra library for this pattern.

src/domain/harbors/HarborsPageContainer.tsx
Zeukkari: Agreed. This is difficult to read and doesn't validate the data structure.

src/domain/harbors/HarborsPageContainer.tsx
midori0507: Well you might somehow injected lodash by some package dependencies (check yarn.lock). And also partial import doesnt hurt as much as full lib import.

Anyway, bad thing about lodash -> swallowed error.

I was 50:50 about using lodash so no is fine

src/domain/harbors/HarborsPageContainer.tsx
```diff
@@ -0,0 +1,127 @@
...
+          return acc;
+        }, [])
+      : [];
+
+  return (
+    <Table
+      data={tableData}
+      columns={columns}
+      renderSubComponent={row => row.index}
+      renderMainHeader={() => t('harbors.tableHeaders.mainHeader')}
```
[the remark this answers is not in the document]
Zeukkari: I was wrong. The table row is contextually bound to the table selected items and render props is probably the nicest approach.

src/domain/harbors/HarborsPageContainer.tsx
Nizar-Rahme: I opted to not use `lodash`. Specifically for this particular issue since optional chaining is in [beta](https://devblogs.microsoft.com/typescript/announcing-typescript-3-7-beta/)  now.

src/common/icon/Icon.tsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
```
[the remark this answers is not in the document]
Nizar-Rahme: Right, I'll move them to the assets folder.
Honestly, I didn't give it much thought, I took this component from `berth-reservation-ui`.

src/domain/harbors/HarborsPageContainer.tsx
Nizar-Rahme: Yes, this array wasn't typed, so I added the type, which allowed me to get rid of these repeated inline types `{ cell: { value: number } }`.
But other than that and after trying a couple of ways, I figured that things would only get more complicated since I still have to list all the values above.
Let me know if you have another solution!

PR #6, opened 2019-10-24
VEN-227 | Layout component

Nizar-Rahme:


src/assets/styles/_breakpoints.scss
```diff
@@ -0,0 +1,49 @@
...
+    padding: 20px;
+  }
+}
+*/
+
+@mixin for-size($range) {
+  $phone-upper-boundary: 600px;
+  $tablet-portrait-upper-boundary: 900px;
+  $tablet-landscape-upper-boundary: 1200px;
+  $desktop-upper-boundary: 1800px;
```
midori0507: where did you get those value from? they didnt looks so familiar..

src/assets/styles/_spacings.scss
```diff
@@ -19,6 +19,16 @@ $project-spacing-standard: (
...
+/*
+===================================
+    Usage
+===================================
+.element {
+  margin: units(1) units(3);
+  padding: units(2);
+}
+*/
+
```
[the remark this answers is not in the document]
Nizar-Rahme: Actually I have just added it, I thought it would be cool if an example of usage.

src/common/layout/layout.module.scss
```diff
@@ -0,0 +1,41 @@
...
+.layout {
+  display: grid;
+  grid-template-columns: minmax(units(16), 1fr) 4fr;
+  grid-template-rows: minmax(units(4), min-content) 1fr auto;
+  grid-template-areas:
+    'header  header'
+    'side   content'
+    'footer  footer';
+
+  min-height: 100vh;
```
midori0507: i prefer 100% height inherit from html

src/assets/styles/_breakpoints.scss
midori0507: i suggest taking some bootstrap one 😆 something like 768, 1440..

PR #7, opened 2019-10-25
DEVOPS-7 | Docker setup

Zeukkari:
# Docker setup for development

During development, the Docker image needs to be rebuilt after changes to package.json.

Building the image:

```
docker-compose up --build
```

Running the container:

```
docker-compose up
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,8 @@ ARG NODE_ENV=development
 ENV NODE_ENV $NODE_ENV
 
 # Global npm deps in a non-root user directory
-ENV NPM_CONFIG_PREFIX=/home/appuser/.npm-global
-ENV PATH=$PATH:/home/appuser/.npm-global/bin
+ENV NPM_CONFIG_PREFIX=/app/.npm-global
+ENV PATH=$PATH:/app/.npm-global/bin
 
 # Yarn
 ENV YARN_VERSION 1.19.1
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,13 +2,20 @@
 FROM helsinkitest/node:12-slim as appbase
 # ===============================================
 
-COPY yarn.lock /app/yarn.lock
-COPY package.json /app/package.json
+# Create app directory
+WORKDIR /app
+
+# Bundle app source
+COPY . .
+
 RUN yarn
 
 # =============================
 FROM appbase as development
 # =============================
+
+ARG NODE_ENV=development
+ENV NODE_ENV=${NODE_ENV}
 
 COPY . /app
 CMD ["yarn", "start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,19 @@ COPY package*.json *yarn* ./
 
 # Install npm depepndencies
 ENV PATH /app/node_modules/.bin:$PATH
+# RUN yarn && yarn cache clean --force
+
+USER root
+
+RUN apt-install.sh build-essential
+
+USER appuser
 RUN yarn && yarn cache clean --force
+
+USER root
+RUN apt-cleanup.sh build-essential
+
+USER appuser
 
 # =============================
 FROM appbase as development
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,48 +4,35 @@ FROM helsinkitest/node:12-slim as appbase
 # Offical image has npm log verbosity as info. More info - https://github.com/nodejs/docker-node#verbosity
 ENV NPM_CONFIG_LOGLEVEL warn
 
-# We will use the `node` user which comes with the official node alpine image
-ENV HOME /home/node
-
 # set our node environment, either development or production
 # defaults to production, compose overrides this to development on build and run
-ARG NODE_ENV=production
+ARG NODE_ENV=development
 ENV NODE_ENV $NODE_ENV
 
 # Global npm deps in a non-root user directory
-ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
-ENV PATH=$PATH:/home/node/.npm-global/bin
+ENV NPM_CONFIG_PREFIX=/home/appuser/.npm-global
+ENV PATH=$PATH:/home/appuser/.npm-global/bin
 
 # Yarn
 ENV YARN_VERSION 1.19.1
 RUN yarn policies set-version $YARN_VERSION
 
-# UID and GID mapping
-RUN groupmod -g 999 node && usermod -u 999 -g 999 node
-
-# Create a node_modules directory and set it's permissions
-RUN mkdir -p $HOME/app/node_modules && chown -R node:node $HOME/app
-
-# Create app directory
-WORKDIR $HOME/app
-
 # Use non-root user
-USER node
+USER appuser
 
 # Copy package.json and package-lock.json/yarn.lock files
 COPY package*.json *yarn* ./
 
 # Install npm depepndencies
+ENV PATH /app/node_modules/.bin:$PATH
 RUN yarn && yarn cache clean --force
-ENV PATH $HOME/node_modules/.bin:$PATH
 
 # =============================
 FROM appbase as development
 # =============================
 
 # copy in our source code last, as it changes the most
-WORKDIR $HOME/app
-COPY --chown=node:node . .
+COPY --chown=appuser:appuser . .
 
 # Bake package.json start command into the image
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# Yarn
+ENV YARN_VERSION 1.19.1
+RUN yarn policies set-version $YARN_VERSION
+
+# UID and GID mapping
+RUN groupmod -g 999 node && usermod -u 999 -g 999 node
+
+# Create a node_modules directory and set it's permissions
+RUN mkdir -p $HOME/app/node_modules && chown -R node:node $HOME/app
```
frwickst: I have not seen this as being necessary at any point creating the node_modules dir like this. It will be created by  the same user as the one that runs `yarn` anyway, which should be the `appuser`

The CHOWN is not needed as you should be using the `appuser`

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# Create app directory
+WORKDIR $HOME/app
```
frwickst: This is already defined in the base image and can be removed.

Dockerfile
```diff
@@ -0,0 +1,51 @@
+# ===============================================
+FROM helsinkitest/node:12-slim as appbase
+# ===============================================
+# Offical image has npm log verbosity as info. More info - https://github.com/nodejs/docker-node#verbosity
+ENV NPM_CONFIG_LOGLEVEL warn
+
+# We will use the `node` user which comes with the official node alpine image
```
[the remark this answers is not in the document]
frwickst: This env var is also not needed after the changes below have been made.

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+ENV HOME /home/node
+
+# set our node environment, either development or production
+# defaults to production, compose overrides this to development on build and run
+ARG NODE_ENV=production
+ENV NODE_ENV $NODE_ENV
+
+# Global npm deps in a non-root user directory
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
```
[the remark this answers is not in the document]
frwickst: Please use the `/app` dir for all application related code.

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN groupmod -g 999 node && usermod -u 999 -g 999 node
+
+# Create a node_modules directory and set it's permissions
+RUN mkdir -p $HOME/app/node_modules && chown -R node:node $HOME/app
+
+# Create app directory
+WORKDIR $HOME/app
+
+# Use non-root user
+USER node
```
frwickst: Please use `appuser` instead

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+# Copy package.json and package-lock.json/yarn.lock files
+COPY package*.json *yarn* ./
+
+# Install npm depepndencies
+RUN yarn && yarn cache clean --force
+ENV PATH $HOME/node_modules/.bin:$PATH
```
frwickst: Please set this before the `yarn` command as this would create a new Docker layer every time new packages are added. Not a big deal, but a good practise putting things that are run less, higher up in the file to create less layers.

package.json
```diff
@@ -16,7 +16,7 @@
     "classnames": "^2.2.6",
     "graphql": "^14.5.8",
     "i18next": "^17.0.18",
-    "node-sass": "^4.12.0",
+    "node-sass": "^4.13.0",
```
frwickst: Was this needed to create the Dockerfile?

If not, then this change should be in another PR.

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+# =============================
+FROM appbase as development
+# =============================
+
+# copy in our source code last, as it changes the most
+WORKDIR $HOME/app
+COPY --chown=node:node . .
+
+# Bake package.json start command into the image
+CMD ["react-scripts", "start"]
```
frwickst: Here you are running `react-scripts start` but in the docker-compose file you are specifying the command to be `yarn start`. The Dockerimage should be the default here, and since it already specifies that it is the `development` stage, then I would imagine that the docker-compose files `command: yarn start` is redundant and can be removed.

docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+services:
+  app:
+    container_name: berth-reservation-admin
+    build: .
+    volumes:
+      - .:/home/node/app
+      - /home/node/app/node_modules
+    working_dir: /home/node/app
+    environment:
+      NODE_ENV: development
```
frwickst: I would rather like to see this set in the Dockerfile as it has a `development` stage.

docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "3.7"
+services:
+  app:
+    container_name: berth-reservation-admin
+    build: .
+    volumes:
+      - .:/home/node/app
+      - /home/node/app/node_modules
+    working_dir: /home/node/app
```
frwickst: Is there a need for this? I have not needed this myself for any project so far. This should be specified by the WORKDIR in the Dockerfile IMHO.

package.json
Zeukkari: This was actually needed to create the Dockerfile. Something related to node12/node-gyp/node-sass dependency alignment... I'll double check to be sure.

Dockerfile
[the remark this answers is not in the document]
Zeukkari: removed.

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+# Global npm deps in a non-root user directory
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+ENV PATH=$PATH:/home/node/.npm-global/bin
+
+# Yarn
+ENV YARN_VERSION 1.19.1
+RUN yarn policies set-version $YARN_VERSION
+
+# UID and GID mapping
+RUN groupmod -g 999 node && usermod -u 999 -g 999 node
```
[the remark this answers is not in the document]
Zeukkari: removed.

Dockerfile
Zeukkari: check, removed.

Dockerfile
Zeukkari: Check, setting PATH before running yarn now.

Dockerfile
Zeukkari: Removed the `yarn start` from docker-compose file.

docker-compose.yml
Zeukkari: Seems redundant and has been removed.

Dockerfile
Zeukkari: check, I have removed this.

Dockerfile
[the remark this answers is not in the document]
Zeukkari: Check, pointing NPM_CONFIG_PREFIX to `/app/.npm-global`.

docker-compose.yml
Zeukkari: true, I removed this from docker-compose.

package.json
Zeukkari: This actually isn't needed anymore since the container has build-essentials.

Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+USER root
+
+RUN apt-install.sh build-essential
+
+USER appuser
+RUN yarn && yarn cache clean --force
+
+USER root
+RUN apt-cleanup.sh build-essential
```
frwickst: The cleanup step here will not result in much. It will remove things from the final container, but it will not make the image any smaller due to previous steps already being cached.

What would make the image smaller would be something in the lines of:
```
USER root
RUN apt-install.sh build-essential && \
         su - appuser -c "yarn && yarn cache clean --force" && \
         apt-cleanup.sh build-essential
USER appuser
```
I do however agree here that this is not the nicest thing in the word read or write.

Another way could be to install everything as root, and then chown it, that
would make things a bit more readable maybe:
```
USER root
RUN apt-install.sh build-essential && \
         yarn && yarn cache clean --force && \
         apt-cleanup.sh build-essential && \
         sudo chown -R appuser:appuser ./node_modules
USER appuser
```
This has the caveat that if things are _not_ installed to `node_modules` for any reason, then
this will not be enough and will not cover that case.

Zeukkari: Alright, I went with the first option :).

PR #8, opened 2019-10-25
VEN-283 | Set up Storybook

Nizar-Rahme:


src/assets/styles/main.scss
```diff
@@ -1,5 +1,5 @@
 @import 'normalize';
-@import 'colors';
+@import 'colours';
```
midori0507: I think US is better than GB here..
I used to change from favourite to favorite because US

Nizar-Rahme: I did prefer ´colors´ but something weird was going on, ´_colors´ was conflicting with some other ´partial´ from Storybook.

PR #9, opened 2019-10-29
add simple data table to storybook

Zeukkari:
Screenshot from 2019-10-29 10-14-05

### DIFF
--- a/src/common/table/Table.stories.tsx
+++ b/src/common/table/Table.stories.tsx
@@ -1,0 +1,43 @@
+import React from 'react';
+
+import Table from './Table';
+
+export default {
+  component: Table,
+  title: 'Table',
+};
+
+const MOUNTAINS = [
+  { name: 'Kilimanjaro', height: 5895, country: 'Tanzania' },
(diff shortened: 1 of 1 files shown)


src/common/table/Table.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  { name: 'Kilimanjaro', height: 5895, country: 'Tanzania' },
+  { name: 'Everest', height: 8848, country: 'Nepal' },
+  { name: 'Mount Fuji', height: 3776, country: 'Japan' },
+  { name: 'Mont Blanc', height: 4808, country: 'Italy/France' },
+  { name: 'Vaalserberg', height: 323, country: 'Netherlands' },
+  { name: 'Denali', height: 6168, country: 'United States' },
+  { name: 'Popocatepetl', height: 5465, country: 'Mexico' },
+];
+
+const columns: any[] = [
```
Nizar-Rahme: Just wondering, does TS complain if we remove `any[]`?

PR #11, opened 2019-10-29
VEN-250 | Label Value component

Zeukkari:
Screenshot from 2019-10-29 13-14-56


package.json
```diff
@@ -20,6 +20,7 @@
     "react": "^16.10.1",
     "react-dom": "^16.10.1",
     "react-i18next": "^10.13.1",
+    "react-intl": "^3.4.0",
```
Nizar-Rahme: In this project, we're using `react-i18next`, so we don't need this one.

src/common/labelValuePair/labelValuePair.scss
```diff
@@ -0,0 +1,12 @@
+@import 'spacings';
+@import 'colours';
```
Nizar-Rahme: If you are not using any color variable, no need to import this module.

src/common/labelValuePair/LabelValuePair.tsx
```diff
@@ -0,0 +1,21 @@
...
+  value?: string | null;
+}
+
+const LabelValuePair = ({ label, value = '-' }: Props) => (
+  <div className="vene-label-value-pair">
+    <FormattedMessage id={label}>
+      {lbl => <span className="vene-label-value-pair__label">{lbl}</span>}
+    </FormattedMessage>
+    <span className="vene-label-value-pair__label">:</span>
+    <span className="vene-label-value-pair__value">{value}</span>
```
Nizar-Rahme: No need to follow BEM convention for naming classes, since we are using SASS modules.

src/common/labelValuePair/labelValuePair.scss
```diff
@@ -0,0 +1,12 @@
...
+
+.vene-label-value-pair {
+  &__label {
+    font-weight: 500;
+  }
+
+  &__value {
+    margin-left: units(0.5);
+  }
+}
```
Nizar-Rahme: Rename this file to `labelValuePair.module.scss`

src/common/labelValuePair/labelValuePair.scss
Zeukkari: removed

src/common/labelValuePair/labelValuePair.scss
Zeukkari: fixed

src/common/labelValuePair/LabelValuePair.tsx
Zeukkari: fixed

package.json
Zeukkari: removed

Nizar-Rahme: Can you remove it also from the packge.json?

PR #17, opened 2019-11-01
use colors from hds-core

Zeukkari:


Nizar-Rahme:
I like the idea of using HDS color values, but I'm hesitant about spreading their variables all over the application because it will be quite difficult to handle it if breaking changes have been introduced to their API, therefore, I suggest mapping their values to our own in one module and keep using our own.
That's the first approach that came to my mind, maybe we can come up together with a better solution.

PR #23, opened 2019-11-05
VEN-300 | Text component

Nizar-Rahme:


Nizar-Rahme:
@Zeukkari Where did these errors come from? 
Travis build had no issues!

PR #32, opened 2019-11-08
VEN-311 | Feature/git hooks

Zeukkari:
Add pre-commit git hook to catch linter errors early in the CI/CD pipeline.


.huskyrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  hooks: {
+    "pre-commit": "yarn lint"
+  }
+};
```
igordavydsson: newline missin'

Zeukkari: fixed

PR #38, opened 2019-11-08
DEVOPS-36 | alternative hotfix

Zeukkari:


Dockerfile
```diff
@@ -54,6 +54,9 @@ CMD ["react-scripts", "start"]
 FROM appbase as staticbuilder
 # ===================================
 
+ARG REACT_APP_API_URI
+ENV REACT_APP_API_URI $REACT_APP_API_URI
+
```
antti-mikael: Just `ARG REACT_APP_API_URI` should be enough.

Zeukkari: done

PR #45, opened 2019-11-11
VEN-191 | List component

Nizar-Rahme:


src/common/list/List.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import styels from './list.module.scss';
```
Zeukkari: There's a typo here.

src/common/list/List.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface Props {
+  size?: 'small' | 'standard' | 'large';
+  custom?: boolean;
```
Zeukkari: Could we use `className` prop here instead of `custom` as in `Paragraph.tsx`?

src/common/listItem/ListItem.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
Zeukkari: I'm not sure if we should have a separate directory for ListItem component. Why not just keep it grouped with the List component under the same directory?

src/common/list/List.tsx
Nizar-Rahme: Actually, this `custom` prop is not for allowing a custom className; it is a boolean that will strip out the bullet points and all default styles from a list.
Couldn't come up with a better name 😃

Zeukkari: Alright. Would something like `resetStyle` make sense?

PR #47, opened 2019-11-13
VEN-343 | List of current customers

Zeukkari:
- ~~graphql / typescript codegeneration~~
- separate harborlist page into container and representation components
- add storybook story for harborlist
- add dummy customer data for scaffolding
- draft customer list page
Screenshot from 2019-11-13 22-26-50

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,8 +25,3 @@ yarn-error.log*
 storybook-static
 
 /__generated__/
-<<<<<<< HEAD
-
-.vscode/
-=======
->>>>>>> 8c054c9... Fix .gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ yarn-error.log*
 storybook-static
 
 /__generated__/
+
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -25,5 +25,8 @@ yarn-error.log*
 storybook-static
 
 /__generated__/
+<<<<<<< HEAD
 
 .vscode/
+=======
+>>>>>>> 8c054c9... Fix .gitignore
--- a/src/domain/customers/CustomerDetailsComponent.tsx
+++ b/src/domain/customers/CustomerDetailsComponent.tsx
@@ -3,8 +3,6 @@ import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
 
 import Grid from '../../common/grid/Grid';
-import ExternalLink from '../../common/externalLink/ExternalLink';
-import LabelValuePair from '../../common/labelValuePair/LabelValuePair';
 import styles from './customerList.module.scss';
 import Paragraph from '../../common/paragraph/Paragraph';
 
@@ -22,11 +20,7 @@ export interface CustomerData {
   phone?: string;
 }
 
-interface Props {
-  data: any;
-}
-
-const CustomerDetails: React.SFC<Props> = ({ data }) => {
+const CustomerDetails: React.SFC<CustomerData | any> = ({ data }: any) => {
   const { t } = useTranslation();
 
   return (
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -25,3 +25,5 @@ yarn-error.log*
 storybook-static
 
 /__generated__/
+
+.vscode/
```
Nizar-Rahme: I think it's good to keep the common settings shared among team members.
As far as I know, it's still possible that one can customize his/her settings using `user settings`.

src/domain/customers/CustomerDetailsComponent.tsx
```diff
@@ -0,0 +1,43 @@
...
+  id: string;
+  image?: string;
+  invoicingType?: string;
+  language?: string;
+  lastName?: string;
+  nickname?: string;
+  phone?: string;
+}
+
+const CustomerDetails: React.SFC<CustomerData | any> = ({ data }: any) => {
```
Nizar-Rahme: Why do you need `any`? It's better to avoid it as possible.

src/domain/customers/CustomerListComponent.tsx
```diff
@@ -0,0 +1,97 @@
...
+      accessor: 'thing',
+    },
+    {
+      Header: 'Laskuja',
+      accessor: 'invoice',
+    },
+    {
+      Header: 'Avaa',
+      accessor: 'goToDetails',
+    },
```
Nizar-Rahme: Header's value should be added to `fi.json`and retrieved from it.

src/domain/customers/CustomerListComponent.tsx
```diff
@@ -0,0 +1,97 @@
...
+  group: string;
+  invoice: string;
+  name: string;
+  queue: string;
+  startDate: string;
+  thing: string;
+}
+
+interface Props {
+  data: CustomerData | null | any;
```
Nizar-Rahme: Also here, better to avoid `any`, since it just suppresses TypeScript errors.

.gitignore
Zeukkari: I decided to install a vscode extension that prrovides unique theme for each workspace. I didn't want to enforce this so I put vscode settings to gitognore. You have a fair point through. What do you think?

Zeukkari: https://marketplace.visualstudio.com/items?itemName=stuart.unique-window-colors

src/domain/customers/CustomerDetailsComponent.tsx
Zeukkari: I dont have access to type definitions yet. Any types shall be removed in another PR.

src/domain/customers/CustomerListComponent.tsx
Zeukkari: same as above

src/domain/customers/CustomerListComponent.tsx
Zeukkari: fixed

.gitignore
Zeukkari: Problem solved with a global gitignore file.

.gitignore
```diff
@@ -25,3 +25,8 @@ yarn-error.log*
 storybook-static
 
 /__generated__/
+<<<<<<< HEAD
+
+.vscode/
+=======
+>>>>>>> 8c054c9... Fix .gitignore
```
Nizar-Rahme: Can you fix this, please?

src/domain/customers/CustomerDetailsComponent.tsx
Nizar-Rahme: Can you leave a `TODO` comment for fixing it?

.gitignore
Zeukkari: fixed

PR #49, opened 2019-11-21
fix graphql type generation

Zeukkari:


package.json
```diff
@@ -14,8 +14,10 @@
     "@types/react-dom": "16.9.1",
     "@types/react-router-dom": "^5.1.0",
     "apollo-boost": "^0.4.4",
+    "apollo-client": "^2.6.4",
     "classnames": "^2.2.6",
     "graphql": "^14.5.8",
+    "graphql-tag": "^2.10.1",
```
Nizar-Rahme: We have `apollo-boost` installed which [includes](https://github.com/apollographql/apollo-client/tree/master/packages/apollo-boost#whats-in-apollo-boost) both `apollo-client` and `graphql-tag`.

Zeukkari: Unnecessary dependency removed

PR #50, opened 2019-11-21
VEN-323 | Individual Harbor Page

Nizar-Rahme:


Nizar-Rahme:
Fixed the requested changes, though I left Page under the `domain` folder since it's only relevant to our service. 

> **Domain:** Domain files are the project specific files that are tied to the projects business logic.
**Common:** Common files are the files that aren't tied to the projects business logic. Basically, you could share these files between projects, and even maybe make them as npm packages. Though, this is not a requirement for a common file.

At least that's what I understood from [frontend folder structure docs](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/DD/pages/786510/Frontend+folder+structure) 😃

Nizar-Rahme:
I have wrapped all routes in the `Page` component as you suggested in #51.
I don't think `Card` will come in handy for this component, otherwise, it'll require more customization than the wrapper of this component deserves.

PR #51, opened 2019-11-21
refactor

Zeukkari:
Refactor Page container?

### DIFF
--- a/src/domain/app/App.tsx
+++ b/src/domain/app/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
 
+import Page from '../page/Page';
 import HarborsPage from '../harbors/HarborsPageContainer';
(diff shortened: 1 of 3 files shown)


src/domain/app/App.tsx
```diff
@@ -1,16 +1,19 @@
...
 import IndividualHarborPage from '../individualHarbor/IndividualHarborPageContainer';
 
 const App: React.FC = () => {
   return (
     <Router>
       <Switch>
-        <Route exact path="/" component={HarborsPage} />
-        <Route path="/harbors/:id" component={IndividualHarborPage} />
-        <Route path="/harbors" component={HarborsPage} />
+        <Page>
```
Nizar-Rahme: Yeah, I like this proposal, the only comment I have is that `Switch` should be moved under `Page` to wrap `Routes`.
Could you push it after fixing this thing to https://github.com/City-of-Helsinki/berth-reservations-admin/pull/50?

PR #53, opened 2019-11-27
VEN-343 | List of current customers 

Zeukkari:
- Add customer list to sidebar


src/common/internalLink/InternalLink.tsx
```diff
@@ -7,10 +7,12 @@ import styles from './internalLink.module.scss';
 type Props = {
   color?: 'standard' | 'critical' | 'brand';
   underlined?: boolean;
+  icon?: JSX.Element;
 } & LinkProps;
 
 const InternalLink: React.SFC<Props> = ({
   color = 'brand',
+  icon,
```
Nizar-Rahme: I guess the icon here is redundant since it is not rendered.

Zeukkari: it's rendered in the sidebar

src/domain/page/Page.tsx
```diff
@@ -22,12 +23,30 @@ const Page: React.SFC = ({ children }) => {
...
-                </InternalNavLink>
+                <InternalLink to="harbors">
+                  <Button
+                    variant="text"
+                    icon={<Icon name="fence" color="standard" />}
+                    onClick={() => {}}
+                  >
+                    <Text bold>Venepaikat</Text>
+                  </Button>
+                </InternalLink>
```
Nizar-Rahme: I do prefer giving the button an `onClick` prop that takes care of the navigation like: `() => history.push('harbors');`
rather than wrapping the whole button by a link and giving the button a dummy function.

src/domain/page/Page.tsx
```diff
@@ -22,12 +23,30 @@ const Page: React.SFC = ({ children }) => {
...
+              key="customers"
+              onClick={() => history.push('/customers')}
+              label={
+                <InternalLink to="customers">
+                  <Button
+                    variant="text"
+                    icon={<Icon name="individual" color="standard" />}
+                    onClick={() => {}}
+                  >
+                    <Text bold>Asiakkaat</Text>
```
Nizar-Rahme: Can you move the text here to the `fi.json` file?

src/common/internalLink/InternalLink.tsx
Nizar-Rahme: I mean here in the return value of this InternalLink component it is not used anywhere! 
```jsx
return (
    <Link
      {...props}
      className={classNames(styles.internalLink, styles[color], {
        [styles.underlined]: underlined,
      })}
    >
      {children}
    </Link>
  );
```

Zeukkari: You're right :). Fixed.

src/domain/page/Page.tsx
Zeukkari: Yeah, I also added some other missing translations.

src/domain/page/Page.tsx
Zeukkari: HarborsContainer handles navigation without history so I decided to remove it for now. I also removed the dummy functions since they were redundant.

PR #55, opened 2019-11-27
Refactoring

Zeukkari:
Remove react-router-dom dependency from header and helsinki logo -components.


src/common/externalLink/ExternalLink.tsx
```diff
@@ -7,7 +7,7 @@ export interface Props {
   to?: string;
   underline?: 'none' | 'hover' | 'always';
   variant?: 'default' | 'withArrow';
-  children: React.ReactNode;
+  children?: React.ReactNode;
```
Nizar-Rahme: `ReactNode` includes `null`s and `undefined`, so no need to make it optional.

src/common/header/Header.tsx
```diff
@@ -1,12 +1,15 @@
 import React from 'react';
 
+import ExternalLink, { Props as LinkProps } from '../externalLink/ExternalLink';
 import HelsinkiLogo from '../helsinkiLogo/HelsinkiLogo';
 import styles from './header.module.scss';
 
-const Header: React.SFC = () => {
+const Header = ({ href }: LinkProps) => {
```
Nizar-Rahme:
```suggestion
const Header: React.SFC<LinkProps> = ({ href }) => {
```

src/common/header/Header.tsx
```diff
@@ -1,12 +1,15 @@
...
   return (
     <header className={styles.header}>
-      <HelsinkiLogo color="white" />
+      <ExternalLink href={href}>
```
Nizar-Rahme: This shouldn't be an external link.
Unlike `react-router-dom` links, normal (external) links force the page to refresh.

Zeukkari: do we really need routing here?

Nizar-Rahme: I would say yes 🙂

PR #56, opened 2019-11-28
VEN-320 - Berths Table on Individual Harbour Page

Nizar-Rahme:


src/common/utils/utils.ts
```diff
@@ -0,0 +1,7 @@
+export const covertToString = (value: unknown, fallback = '') => {
```
Zeukkari: Please remove these utility functions.

src/index.tsx
```diff
@@ -17,6 +17,12 @@ const {
 } = process.env;
 
 const client = new ApolloClient({
+  request: async operation => {
```
Zeukkari: please move this to another PR

PR #71, opened 2019-12-31
VEN-290 | Use colors from hds-core and add missing icons

Zeukkari:
Screenshot from 2019-12-31 16-48-44


Zeukkari:
This branch uses an older version of hds-core and it's not worth maintaining anymore. Sidebar changes are included in https://github.com/City-of-Helsinki/berth-reservations-admin/pull/87.

PR #73, opened 2020-01-08
VEN-361 | Add git commit hook for graphql codegeneration

Zeukkari:
Update apollo remote schema and generate typings during git commits to catch errors caused by changed schema.

Possible future development: I'm considering switching apollo's codegenerator to [graphql-code-generator](https://graphql-code-generator.com/) as it seems to be much more flexible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "deploy-storybook": "storybook-to-ghpages",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -87,4 +88,4 @@
     "husky": "^3.0.9",
     "prettier": "^1.18.2"
   }
-}
+}
```
igordavydsson: newline missing 🙃

PR #88, opened 2020-01-21
VEN-337 | Customer's application card

Nizar-Rahme:


src/locales/fi.json
```diff
@@ -98,6 +98,26 @@
...
+      "boatType": "Veneen tyyppi",
+      "registrationNumber": "Rekisterinumero",
+      "boatWidth": "Veneen leveys",
+      "boatLength": "Veneen pituus",
+      "boatDepth": "Veneen syväys",
+      "boatWeight": "Veneen paino",
+      "boatName": "Nimi",
+      "boatBrand": "Merkki",
+      "selectedPorts": "VALITUT SATAMAT",
+      "choice": "Toive {{num}}",
```
Zeukkari: Can we move this template string to the application component?

src/common/checkbox/Checkbox.tsx
```diff
@@ -4,23 +4,37 @@ import classNames from 'classnames';
 import styles from './checkbox.module.scss';
 import Icon from '../icon/Icon';
```
Zeukkari: Can we switch to Icon component from HDS?

Nizar-Rahme: The [Icon component](https://city-of-helsinki.github.io/helsinki-design-system/storybook/?path=/story/icons--iconcheck) from HDS won't work for this checkbox, but I can replace the Icon here with an imported SVG, so the checkbox would be a self-contained component as we discussed.
What do you think?

src/domain/individualCustomer/billsCard/BillsCard.test.jsx
```diff
@@ -25,7 +25,7 @@ describe('BillsCard', () => {
   });
   const getWrapper = (props = mockProps) => shallow(<BillsCard {...props} />);
 
-  it('renders normally', () => {
+  it('renders correctly', () => {
```
Zeukkari: Can we revert this?

PR #89, opened 2020-01-24
VEN-334 | BoatsCard on individual customer page

Nizar-Rahme:


src/domain/individualCustomer/boatsCard/BoatsCard.tsx
```diff
@@ -0,0 +1,118 @@
...
+  boats: (Boat | LargeBoat)[];
+}
+
+const BoatsCard: React.SFC<BoatsCardProps> = ({ boats }) => {
+  const { t } = useTranslation();
+  const isLargeBoat = (boat: LargeBoat | Boat): boat is LargeBoat =>
+    (boat as LargeBoat).boatPower !== undefined;
+
+  return (
+    <div>
```
Zeukkari: I think this `div` should be a `Card` and the nested `Card` under this should be a `Paragraph`.

PR #91, opened 2020-01-27
VEN-219 | Feature/refine logout

Zeukkari:


src/domain/pageHeader/PageHeaderContainer.tsx
```diff
@@ -29,7 +29,15 @@ const PageHeaderContainer: React.SFC = () => {
       <Dropdown label={<Button icon={<Icon name="user" />}>{fullName}</Button>}>
         <List noBullets>
           <ListItem>
-            <Button variant="text" onClick={() => dispatch(userSignedOut())}>
+            <Button
+              variant="text"
+              onClick={() => {
+                // Flush data in redux store and localStorage
+                persistor.flush();
```
Nizar-Rahme: Are we really persisting the data in the first place? I think we are not.

src/domain/auth/authenticate.ts
```diff
@@ -50,6 +48,19 @@ export const authenticateWithBackend = (
 
     dispatch(fetchTokenSuccess(res.data));
   } catch (error) {
-    dispatch(fetchTokenError(error));
+    toast(i18n.t('authentication.errorMessage'), {
+      type: toast.TYPE.ERROR,
+    });
+    try {
+      // This is a workaround that can save us until we can fix silentRenew().
```
Nizar-Rahme: Would be better great if we create a ticket in the balcklog for this fix and mention it here, so it won't slip through the cracks.

package.json
```diff
@@ -31,6 +31,7 @@
     "react-router-dom": "^5.1.2",
     "react-scripts": "^3.3.0",
     "react-table": "^7.0.0-rc.15",
+    "react-toastify": "^5.5.0",
```
Nizar-Rahme: I bet it is a temporary solution until we implement the notifications Joakim has designed, isn't it?

src/domain/auth/authenticate.ts
```diff
@@ -50,6 +48,19 @@ export const authenticateWithBackend = (
 
     dispatch(fetchTokenSuccess(res.data));
   } catch (error) {
-    dispatch(fetchTokenError(error));
+    toast(i18n.t('authentication.errorMessage'), {
```
Nizar-Rahme: You forgot to add this error message to `fi.json` file.

src/domain/auth/authenticate.ts
Zeukkari: we already have it - VEN-378

Nizar-Rahme: Cool, so can you mention it in the comment?

src/domain/auth/authenticate.ts
Zeukkari: Fixed :)

package.json
Zeukkari: Yeah, this is analogous to the spinning loader we added last week. Let's replace it with our own implementation when we have the time.

src/domain/pageHeader/PageHeaderContainer.tsx
Zeukkari: This is an intermediary solution which I would like to refactor in the long run. For now, we are using `redux-persist` package. I haven't properly tested this and would like to get rid of this all together.

Nizar-Rahme: I looked into it, it's safe to say that we can get rid of it now (unless we opt to persist the login state). I believe this `redux-persist` has leaked to our code from another service, and it is not used at the moment.

package.json
Nizar-Rahme: Fair enough.

src/domain/pageHeader/PageHeaderContainer.tsx
Zeukkari: ok

src/domain/auth/authenticate.ts
Zeukkari: done

PR #95, opened 2020-01-28
VEN-415 | use fonts from Helsinki Design System

Zeukkari:
Screenshot from 2020-01-28 13-01-28

As far as I understand the fonts are served as files with a .woff extension. I can't seem find the fonts anywhere in hds-core. My current approach is to navigate to https://city-of-helsinki.github.io/helsinki-design-system/storybook and copy the font url from there.

Umm, is this the right approach?

### DIFF
--- a/src/assets/styles/main.scss
+++ b/src/assets/styles/main.scss
@@ -1,12 +1,15 @@
+@import '~hds-core/lib/helsinki.css';
 @import 'normalize';
 @import 'boxSizing';
 @import 'colours';
 
-body {
-  margin: 0;
-  font-family: HelsinkiGrotesk, Helvetica, Arial, 'MS Gothic', Meiryo, Osaka;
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -1,5 +1,6 @@
-import './assets/styles/main.scss';
+// import './assets/styles/main.scss';
```
Nizar-Rahme: Why do you have to comment this out?

Zeukkari: It didn't seem to do anything and I considered global styles should come from hds-core.

Zeukkari: I uncommented that file and decided to move font face definitions there. However, in the long run I'd like to get rid of the `assets/` folder entirely.

Zeukkari:
I refined this PR so that Helsinki Grotesk font is now used globally instead of the button component.

PR #100, opened 2020-02-03
VEN-431 | Feature/deprecate button

Zeukkari:
- Deprecates our Button component in favor of HDS
- Deprecates our Icon component in favor of HDS
- Adds new components IconOutline and NavLink
- This PR could be cleaned up a bit. The SVG files could be removed and IconOutline component is defined a couple of times.

Screenshot from 2020-02-06 17-26-32


Nizar-Rahme:
image
These new buttons don't look right!

src/common/internalLink/NavLink.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <button
+        type="button"
+        onClick={onClick}
+        className={classNames(
+          styles.button,
+          styles[color],
+          styles.hasIcon,
+          className
+        )}
+      >
```
Nizar-Rahme: Why do we need a button here inside the link?

src/common/internalLink/internalLink.module.scss
```diff
@@ -5,6 +5,13 @@
 
   text-decoration: none;
   color: var(--vene-internal-link-color);
+
+  display: flex;
+  align-items: center;
+
+  & > :first-child {
+    margin-right: 1rem;
```
Nizar-Rahme: Can you use the spacing units we're using in other places to keep the interface consistent?

Zeukkari:
> Is this PR for deprecating the Button **and** the Icon?

Yes. Please see VEN-450 for further information.

src/common/internalLink/NavLink.tsx
Zeukkari: We don't, button has been removed.

src/common/internalLink/internalLink.module.scss
Zeukkari: using spacing units here

Zeukkari:
Merging this now. I split leftover work into VEN-456 and VEN-457.

PR #124, opened 2020-02-27
VEN-481 | Remove the Radio Button from Table's Header

Nizar-Rahme:


src/common/checkbox/checkbox.module.scss
```diff
@@ -2,10 +2,10 @@
 @import 'colours';
 
 .checkbox {
-  --checkbox-dimension: #{units(1)};
+  --checkbox-dimension: 20px;
```
midori0507: are you sure you want those free values after all HDS tokens and variables and stuff?

Nizar-Rahme: Yeah, I stick to these predefined values for margins, paddings, etc. 
But for the checkbox and radio dimensions, I think these custom values look better.

PR #134, opened 2020-03-30
VEN-515 | Authentication flow

Nizar-Rahme:
## Description :sparkles:
- Remove redux and its dependencies
- Implement Auth service based on `oidc-client`
- Persist the token in the local storage
- Remove the local schema
- Remove redundant components and files

### Automated tests :gear:️
- https://github.com/City-of-Helsinki/berth-reservations-admin/blob/cd390ec4f6f9406ea1f53dd1ecf13221ad15a4c4/src/domain/auth/authService.ts#L1

- Click on the login button which redirects you to tunnistamo 
- After logging in tunnistamo should redirects you back to the page you were trying to access in the first place or the main page i.e `/harbors`
- Refresh the page manually and you should remain logged in
- On the header of the page, your name, fetched from the service you logged in through, appears on the right side
- Click on your name which displays a dropdown with a logout button
- Click on the logout button and you should be logged out

Advanced steps:
- Clear the local storage manually
- Try to navigate to any page the service should log you out and redirects you to the login page


src/domain/auth/authService.ts
```diff
@@ -0,0 +1,113 @@
...
+    });
+  }
+
+  public renewToken(): Promise<User> {
+    return this.userManager.signinSilent();
+  }
+
+  public endRenewToken(): Promise<User | undefined> {
+    return this.userManager.signinSilentCallback();
+  }
```
TatuArvela: I was slightly confused by the `login`&`endLogin` naming, but in the end I'm not sure what else to suggest

TatuArvela:
I went through the code and tested this locally. The quality of the new authentication flow is good, and it appears to work as expected. LGTM :thumbsup:

src/domain/auth/authService.ts
Nizar-Rahme: I borrowed `end` prefix from `oidc-client` [samples](https://github.com/IdentityModel/oidc-client-js/blob/c79d65184a369fbccd3b87fc0ba3f7824554c62c/samples/VanillaJS/public/user-manager-sample.js#L113) for the lack of a better name. 🤭

PR #158, opened 2020-04-17
VEN-557 | Update company info fetching logic

TatuArvela:
## Description :sparkles:

Update company info fetching logic (company -> organization)

### Closes :no_good_woman:

[VEN-557](https://helsinkisolutionoffice.atlassian.net/browse/VEN-557)

No visual changes, merely a data query change. Single application page should work with this.

### DIFF
--- a/src/domain/individualApplication/IndividualApplicationPage.tsx
+++ b/src/domain/individualApplication/IndividualApplicationPage.tsx
@@ -25,7 +25,7 @@ export enum CUSTOMER_GROUP {
   COMPANY = 'COMPANY',
   INTERNAL = 'INTERNAL',
   NON_BILLABLE = 'NON_BILLABLE',
-  OTHER = 'OTHER',
+  OTHER_ORGANIZATION = 'OTHER_ORGANIZATION',
 }
 
 export interface CustomerData {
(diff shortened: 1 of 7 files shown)


src/domain/individualApplication/IndividualApplicationPage.tsx
```diff
@@ -22,7 +22,7 @@ import OfferCard, { OfferCardProps } from './offerCard/OfferCard';
 
 export enum CUSTOMER_GROUP {
   PRIVATE = 'PRIVATE',
-  COMPANY = 'COMPANY',
+  ORGANIZATION = 'ORGANIZATION',
```
igordavydsson, line 25: if it is `organization`, please check what type and use that as `CUSTOMER_GROUP` 🙂

src/domain/individualApplication/IndividualApplicationPage.tsx
```diff
@@ -23,6 +23,9 @@ import OfferCard, { OfferCardProps } from './offerCard/OfferCard';
 export enum CUSTOMER_GROUP {
   PRIVATE = 'PRIVATE',
   COMPANY = 'COMPANY',
+  INTERNAL = 'INTERNAL',
+  NON_BILLABLE = 'NON_BILLABLE',
+  OTHER = 'OTHER',
```
igordavydsson, line 28: should this enum member be renamed too?

igordavydsson: my point is that `OTHER` makes sense as `OrganizationType`, since it is "other type of organization". but when it comes to customer groups, here it might make things clearer if we explicitly state this this is `other organization` and not `other customer group` 🙂

PR #160, opened 2020-04-23
VEN-576 | Implement pier selection header for individual harbor table

ahie:
## Description :sparkles:

Adds a header to the individual harbor page's berth table that allows for filtering by pier.

### Closes :no_good_woman:

VEN-576

Open an individual harbor's page. A pier selection header should be present that conforms to the designs and filters the table by selection.


ahie:
As the type of berth number has been changed in the backend this PR also changes the relevant TS type definitions.

Nizar-Rahme:
Imo, it's better to let the table handles the filtering functionality instead of us filtering the data passed to it since we're going to combine multiple filters together later on, and that might make things more complicated.
https://github.com/City-of-Helsinki/berth-reservations-admin/blob/983756640b70e37ac4a1ee287c007650a94d5f33/src/domain/applications/ApplicationsPageContainer.tsx#L164

Currently, if the table doesn't have a main header, the "secondary" header will stick to the top because of this className:
https://github.com/City-of-Helsinki/berth-reservations-admin/blob/983756640b70e37ac4a1ee287c007650a94d5f33/src/common/table/Table.tsx#L298
So if we can control the stickiness of the main header by passing a prop or so and making use of `renderMainHeader`, it would be perfect.

This component might come in handy `src/common/tableFilters/TableFilters.tsx`

ahie:
> Imo, it's better to let the table handles the filtering functionality instead of us filtering the data passed to it since we're going to combine multiple filters together later on, and that might make things more complicated.

Indeed. I now refactored the code to use `setFilter` instead. And the header is added with `renderTableToolsTop` which takes it out of the main header styling flow.

> Currently, if the table doesn't have a main header, the "secondary" header will stick to the top because of this className:
> 
> https://github.com/City-of-Helsinki/berth-reservations-admin/blob/983756640b70e37ac4a1ee287c007650a94d5f33/src/common/table/Table.tsx#L298
> 
> So if we can control the stickiness of the main header by passing a prop or so and making use of `renderMainHeader`, it would be perfect.

Problem is the stickiness of the next row of headers, which is hardcoded to `top: 3rem`. Should adjust that so that the whole block of different headers stays sticky.

> This component might come in handy `src/common/tableFilters/TableFilters.tsx`

Just now noticed that this class exists.

ahie:
Updated snapshots, should fix the build.

PR #163, opened 2020-04-24
VEN-538| Single application page: searching for similar customers

Nizar-Rahme:
## Description :sparkles:
- Add `Select` and `TextField` for searching similar customers on the single application page
- Use the debouncing mechanism for fetching customers while typing  
- Use lazy selector for fetching similar customers to get rid of on unnecessary roundtrip if the application has a customer attached

### Closes :no_good_woman:
[VEN-538](https://helsinkisolutionoffice.atlassian.net/browse/VEN-538): Single application page: filters for the suggested customers

- Go to the application page
- Click on one of the applications
- If the application doesn't have a customer, a table with suggested customers would be displayed
- Above the table, there are a search bar and a dropdown menu can be used to search for customers based on their `first name`, `last name`, `email` or `address`.

## Screenshots :camera_flash:
image

I left out the pagination to be done in another ticket since I couldn't find a non-hacky way to do it.

### DIFF
--- a/src/domain/individualApplication/IndividualApplicationPageContainer.tsx
+++ b/src/domain/individualApplication/IndividualApplicationPageContainer.tsx
@@ -77,24 +77,8 @@ const IndividualCustomerPageContainer: React.SFC = () => {
   const [deleteDraftedApplication] = useDeleteBerthApplication();
   useEffect(() => {
     if (!data?.berthApplication) return;
-    let initialSearchBy: keyof BerthApplication;
 
-    switch (searchBy) {
-      case SearchBy.EMAIL:
-        initialSearchBy = 'email';
-        break;
-      case SearchBy.ADDRESS:
-        initialSearchBy = 'address';
-        break;
-      case SearchBy.FIRST_NAME:
-        initialSearchBy = 'firstName';
-        break;
-      default:
-        initialSearchBy = 'lastName';
-        break;
-    }
-
-    setSearchVal(data.berthApplication[initialSearchBy]);
+    setSearchVal(data.berthApplication[searchBy]);
   }, [data, searchBy]);
 
   useEffect(() => {
--- a/src/domain/individualApplication/IndividualApplicationPageContainer.tsx
+++ b/src/domain/individualApplication/IndividualApplicationPageContainer.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react';
-import { useQuery, useMutation } from '@apollo/react-hooks';
+import { useQuery, useMutation, useLazyQuery } from '@apollo/react-hooks';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router';
 import { Notification } from 'hds-react';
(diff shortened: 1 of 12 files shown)


src/domain/individualApplication/IndividualApplicationPageContainer.tsx
```diff
@@ -173,7 +220,19 @@ const IndividualCustomerPageContainer: React.SFC = () => {
...
       handleLinkCustomer={handleLinkCustomer}
-      handleCreateCustomer={handleCreateCustomer}
+      customerTableTools={{
+        searchVal,
+        searchBy,
+        setSearchVal,
+        setSearchBy,
+        handleCreateCustomer,
+        searchByOptions: [
+          { value: SearchBy.FIRST_NAME, label: 'Etunimi' },
```
ahie, line 230: Labels are not translated here.

src/domain/individualApplication/queries.ts
```diff
@@ -1,11 +1,18 @@
 import { gql } from 'apollo-boost';
 
 export const FILTERED_CUSTOMERS_QUERY = gql`
-  query FILTERED_CUSTOMERS($firstName: String, $lastName: String) {
+  query FILTERED_CUSTOMERS(
+    $firstName: String
+    $lastName: String
+    $emails_Email: String
```
ahie, line 7: >$emails_Email
>$addresses_Address

Would it make sense to change these variables to simply `email` and `address`? Wouldn't need the switch in `IndividualApplicationPageContainer`.

src/domain/individualApplication/IndividualApplicationPageContainer.tsx
```diff
@@ -51,17 +57,58 @@ const IndividualCustomerPageContainer: React.SFC = () => {
...
+        break;
+      case SearchBy.FIRST_NAME:
+        initialSearchBy = 'firstName';
+        break;
+      default:
+        initialSearchBy = 'lastName';
+        break;
+    }
+
+    setSearchVal(data.berthApplication[initialSearchBy]);
```
ahie, line 97: Not really a comment on the code, just a question about the business logic here.

Whose data is being set here as the initial value if the application is not connected to anybody?

src/domain/individualApplication/IndividualApplicationPageContainer.tsx
Nizar-Rahme: Oops

src/domain/individualApplication/queries.ts
Nizar-Rahme: Good point 👍 
Fixed!

src/domain/individualApplication/IndividualApplicationPageContainer.tsx
Nizar-Rahme: Good question! 
If the application doesn't have a customer, it means that it's not yet linked to an existing customer in our database, but still, it has the personal info of the one who submitted it, and we populate the initial values of the search field with this personal info.

ahie: That makes sense. Thanks for the clarification :+1:

PR #164, opened 2020-04-24
VEN-579 | Add global filter to berth table in individual harbor page

ahie:
## Description :sparkles:

Adds a search field that performs global filtering of berths in the individual harbor page's table.

Additionally refactors filtering by piers to not use component state.

### Closes :no_good_woman:

VEN-579

* Go to a harbor's page.
* Type something in the search text field.
* Only the rows with at least one cell matching the written text should be shown.

### DIFF
--- a/src/domain/individualHarbor/IndividualHarborPageContainer.tsx
+++ b/src/domain/individualHarbor/IndividualHarborPageContainer.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import { useParams } from 'react-router-dom';
 import { useQuery } from '@apollo/react-hooks';
 import { useTranslation } from 'react-i18next';
-import { UseGlobalFiltersOptions } from 'react-table';
 
 import Table, { Column } from '../../common/table/Table';
 import { INDIVIDUAL_HARBOR_QUERY } from './queries';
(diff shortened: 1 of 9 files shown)


src/domain/individualHarbor/IndividualHarborPageContainer.tsx
```diff
@@ -26,7 +27,38 @@ const IndividualHarborPageContainer: React.SFC = () => {
...
+          numberMatch ||
+          identifierMatch ||
+          lengthMatch ||
+          widthMatch ||
+          mooringTypeMatch
+        );
+      });
+    },
+    [t, i18n.language]
+  );
```
Nizar-Rahme, line 61: I find it a little bit error-prone to have a separate function for global filtering because with this approach any change to the `Cell` formatted, has to be repeated it for the global function as well, otherwise, it won't work.
But this is just an opinion.

ahie: I though of refactoring the cells' formatters to functions that would also be used by the filter. Would that be an acceptable solution?

Nizar-Rahme: I think that'd be better, but is there any specific reason to not use the `accessor`s' functions?

ahie: We had a short discussion on slack that doing the formatting in the `accessor` would be bad practice. Now that I've taken another look, we either need to accept that the cell formatting and the global filter are kept in sync, or we move the formatting to the accessor.

Nizar-Rahme: Sorry, I wasn't clear about what I meant by formatting (changing styles 🙈). My bad!
For example, if we have a number value which needs a unit, `10 m`, and this value and it has to be in red, the accessor takes care of adding the unit and Cell do the styles: 
```JSX
accessor: ({ width }) => `${width} m`,
Cell: ({ cell }) => <span style={{ color: 'red' }}>{cell.value}</span>,
```

ahie: Ok! In that case, I'll just change the accessors and update the filter accordingly.

Nizar-Rahme: Thanks!

ahie: Done

Nizar-Rahme: Great 👍 
Do we still need to pass `berthTableGlobalFilter`?
I tried to remove it entirely, and the global filter worked as expected.

ahie: Checked that the global filter defaults to [`text`](https://github.com/tannerlinsley/react-table/blob/master/src/filterTypes.js#L1) if nothing is specified. The behavior is the same, except that it is not locale aware on the lower case conversion. For the languages we aim to support that is not an issue.

Also explicitly added the filter types to the individual column definitions in the latest commit.

Will keep the changes to the `Table` component, as other tables might want to configure the global filter.

Nizar-Rahme: 🚀

PR #167, opened 2020-04-29
VEN-588 | Update boat card

santtul:
## Description :sparkles:

- [x] Remove Boat, Large Boat and Application types duplication
- [x] Display `intendedUse` field in BoatCard for large boats
- [x] Display insurance and inspection certificates  in BoatCard for large boats
- [x] Remove boat fixture data duplication
- [x] Update boats card storybook
- [x] Move boats card tests under `__tests__`

### Closes :no_good_woman:
https://helsinkisolutionoffice.atlassian.net/browse/VEN-588

## Testing :alembic:
- Open customer with large boat: http://localhost:3000/customers/UHJvZmlsZU5vZGU6N2QyMGZlNmMtYWMzZC00ZTY0LWJkYzAtYTJiZjg5ODU5ZGU3
- See that intended use and certificate information is displayed as described in the ticket

### Automated tests :gear:️
Unit tests and React component snapshot test.


Nizar-Rahme:
It looks good to me 👍

igordavydsson:
you might want to edit your git config, you aren't showing as the author 🙂

src/domain/individualCustomer/boatsCard/boatsCardUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  return url.split('/').pop();
+};
+
+export const getValidUntilColor = (
+  validUntil: string
+): 'critical' | 'standard' => {
+  const isBeforeToday =
+    new Date(validUntil) < new Date(new Date().toDateString());
```
Nizar-Rahme, line 30: Isn't this enough?
```suggestion
    new Date(validUntil) < new Date();
```

src/domain/individualCustomer/boatsCard/boatsCardUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+export const isLargeBoat = (boat: LargeBoat | Boat): boat is LargeBoat =>
+  (boat as LargeBoat).boatType.id === LARGE_BOAT_ID;
+
+export const getCertificate = (
+  certificates: BoatCert[],
+  certificateType: BoatCertificateType
+): BoatCert => {
+  return certificates.filter(
+    cert => cert.certificateType === certificateType
+  )[0];
```
Nizar-Rahme, line 15:
```suggestion
  return certificates.find(cert => cert.certificateType === certificateType);
```

src/domain/individualCustomer/boatsCard/BoatCertificate.tsx
```diff
@@ -0,0 +1,63 @@
...
+    validUntil,
+    checkedAt,
+    checkedBy,
+  } = certificate;
+  const headerTextKey = `individualCustomer.customerBoats.${certificateType.toLocaleLowerCase()}`;
+  const filename = getFilename(file);
+
+  return (
+    <Section>
+      <div className={styles.boatCertificate}>
```
Nizar-Rahme, line 30: Why don't we add the `className` directly here and get rid of this extra `div`?
```suggestion
    <Section className={styles.boatCertificate}>
```

src/domain/individualCustomer/boatsCard/boatsCardUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export const getFilename = (url: string | null): string | undefined => {
+  if (!url) {
+    return undefined;
+  }
+
+  return url.split('/').pop();
+};
```
Nizar-Rahme, line 24: I wonder if we can just use [Node's `path` module](https://nodejs.org/api/path.html#path_path_basename_path_ext).

`path.basename('https://venepaikka-api.test.hel.ninja/dummy.pdf')`

src/locales/fi.json
```diff
@@ -338,7 +339,9 @@
       "hullMaterial": "Runkomateriaali",
       "purpose": "Käyttötarkoitus",
       "inspection": "Katsastustodistus",
-      "insurance": "Vakuutuskirjaote"
+      "insurance": "Vakuutuskirjaote",
```
Nizar-Rahme, line 342: Please consider removing this if it's no longer used.

src/domain/individualCustomer/boatsCard/boatsCardUtils.ts
santtul: This is for omitting the time in the "before check". With the code you suggested the valid until with current date would resolve as "not valid anymore" and I assume that the date is "inclusive" (it's still valid at the last day). But agree that might raise eyebrows and if we will start doing more date operations we should start using a a proper date library for that.

src/domain/individualCustomer/boatsCard/BoatCertificate.tsx
santtul: That is a good idea, but it seems that section is generating an `article` sub element so the `display: flex` from `styles.boatCertificate` is then applied to only that and not to the child elements which breaks the layout then.

src/locales/fi.json
santtul: It's still in use.

src/domain/individualCustomer/boatsCard/boatsCardUtils.ts
santtul: We would need to install this dependency:
https://github.com/browserify/path-browserify
Maybe not worth for this small thing? :)

src/domain/individualCustomer/boatsCard/boatsCardUtils.ts
santtul: Fixed.

src/domain/individualCustomer/boatsCard/BoatCertificate.tsx
Nizar-Rahme: Fair enough!

PR #178, opened 2020-05-11
VEN-577, VEN-608 | Implement pier forms

ahie:
## Description :sparkles:

Adds forms for creation, deletion and updating of piers.

### Closes :no_good_woman:

- VEN-577
- VEN-608


src/domain/individualHarbor/forms/pierForm/PierEditForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const [deletePier] = useMutation<DELETE_PIER, DELETE_PIER_VARS>(
+    DELETE_PIER_MUTATION,
+    {
+      refetchQueries: refetchQueries ?? [],
+    }
+  );
+  const { t } = useTranslation();
+
+  if (loading) return <LoadingSpinner isLoading={loading} />;
+  if (error) return <div>{t('forms.common.error')}</div>;
```
Nizar-Rahme, line 55: Should we take into account mutations' errors?

src/domain/individualHarbor/forms/pierForm/PierEditForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+          onDelete?.(values)
+        )
+      }
+      isSubmitting={isSubmitting}
+      suitableBoatTypeOptions={suitableBoatTypeOptions}
+    />
+  );
+};
+
+export default PierEditForm;
```
Nizar-Rahme, line 81: Not sure if we already have a pattern for these forms, but to me `PierFormContainer` sounds more descriptive.

src/domain/individualHarbor/pierSelectHeader/PierSelectHeader.tsx
```diff
@@ -11,40 +11,52 @@ interface PierSelectHeaderProps {
...
+              [styles.pierButtonSelected]: !selectedPier,
             })}
           >
-            {`${t('individualHarbor.tableHeaders.identifier')} ${
-              pier.identifier
-            }`}
+            {t('individualHarbor.tableHeaders.all')}
           </button>
-        ))}
+          {piers.map((pier, idx) => (
```
Nizar-Rahme, line 37: If the pier has an `Id`, it's better to use it as a `key` value instead of the index.

src/domain/individualHarbor/forms/pierForm/PierEditForm.tsx
ahie: `XForm` is the component that contains the actual formik part and the props should extend the interface `FormProps<X>`. That can be used as a base for creating forms with different capabilities, for example `XEditForm` and `XCreateForm`. This pattern has also been used with the berth related forms. Ofcourse this could be refactored to a different approach, suggestions?

src/domain/individualHarbor/forms/pierForm/PierEditForm.tsx
ahie: Added the mutations' errors to this "catch-all" if. Could create a separate issue for creating a more descriptive error component for form errors?

src/domain/individualHarbor/pierSelectHeader/PierSelectHeader.tsx
ahie: Done.

src/domain/individualHarbor/forms/pierForm/PierEditForm.tsx
Nizar-Rahme: Great! Ville might have something in mind regarding these errors.

src/domain/individualHarbor/forms/pierForm/PierEditForm.tsx
Nizar-Rahme: Fair enough!
As long as we're following the same naming pattern in other forms, no need to refactor it; we can continue with that.

PR #179, opened 2020-05-12
VEN-618 | Store the new user in localStorage before trying to access the app

Nizar-Rahme:
## Description :sparkles:
- Await the new tokens to be fetched and the new user to be stored in `localStorage`
- Clear `localStorage` when the user logs out
- Writing unit tests

### Closes :no_good_woman:
[VEN-618](https://helsinkisolutionoffice.atlassian.net/browse/VEN-618): Login requires manual refresh or a second attempt to access the application 

### Related :handshake:
This also should fix the [error in fetching the data in the initial request](https://helsinkicity.slack.com/archives/CBA1QHTHN/p1589272220364600)

### Automated tests :gear:️
https://github.com/City-of-Helsinki/berth-reservations-admin/blob/1ff303da090a47b40419c9712a24b0721973f278/src/domain/auth/authService.test.js#L161
https://github.com/City-of-Helsinki/berth-reservations-admin/blob/1ff303da090a47b40419c9712a24b0721973f278/src/domain/auth/authService.test.js#L215

- Log out
- Log in
- The user should be redirected back to the front page (harbors) instead of the application's login screen


src/domain/auth/authService.ts
```diff
@@ -79,22 +79,20 @@ class AuthService {
...
   public renewToken(): Promise<User> {
     return this.userManager.signinSilent();
   }
 
-  public logout(): Promise<void> {
-    return this.userManager.signoutRedirect().then(() => {
-      localStorage.removeItem(API_TOKENS);
-    });
+  public async logout(): Promise<void> {
+    await this.userManager.signoutRedirect();
```
santtul, line 93: I think there is still a possibility for a race condition here. `signoutRedirect` will redirect to another url. So if the redirection is fast enough the JS in our site may not be executed. Could we change the sequence as follows:
```suggestion
    localStorage.removeItem(API_TOKENS);
    this.userManager.clearStaleState();
    await this.userManager.signoutRedirect();
```

Nizar-Rahme: Good point, fixed!

PR #182, opened 2020-05-14
VEN-575 | Single harbor page: edit harbor properties

TatuArvela:
## Description :sparkles:

* Single harbor page: edit harbor properties
* Added Edit button to HarborProperties header

### Side-effects

* Created new FileUpload component
* Added children "widgets" functionality to CardHeader
* Revealed 'critical' color from Icon
* Adjusted Modal sizing and auto scroll overflow
* Created byte formatter

### Closes :no_good_woman:

* [VEN-575](https://helsinkisolutionoffice.atlassian.net/browse/VEN-575) | Single harbor page: edit harbor properties

### Automated tests :gear:️

* Automated unit tests exist for HarborEditForm, FileUpload and formatBytes
* Updated snapshots for changed CardHeader

* Single harbor page: Click edit
  * Test file upload field


src/common/modal/modal.module.scss
```diff
@@ -24,4 +24,8 @@
   &:focus {
     outline: none;
   }
+  width: 100%;
```
santtul, line 27: Style has no effect

src/common/modal/modal.module.scss
```diff
@@ -24,4 +24,8 @@
...
+  max-width: units(32);
```
santtul, line 28: I think this too restrictive as "global style", for example in ticket https://helsinkisolutionoffice.atlassian.net/browse/VEN-625 I need a wider modal than 32 units.

src/domain/individualHarbor/harborProperties/HarborProperties.tsx
```diff
@@ -28,16 +28,30 @@ export interface HarborPropertiesProps {
...
   const { t } = useTranslation();
 
   const serviceMapUrl = `${process.env.REACT_APP_SERVICE_MAP_URI}${servicemapId}`;
 
   return (
     <Card>
-      <CardHeader title={t('individualHarbor.harborProperties.title')} />
+      <CardHeader title={t('individualHarbor.harborProperties.title')}>
+        <button onClick={editHarbor}>
+          <Text size="m" weight="normalWeight">
```
santtul, line 50: `size="m" ` has no effect

src/common/modal/modal.module.scss
TatuArvela: Actually it does, this makes the modal grow to the given `max-width`, and also makes it scale down on smaller screen sizes (though our responsive design is still practically non-existent)

src/common/modal/modal.module.scss
TatuArvela: That's a good point, maybe the modal should have a size prop for this?

src/domain/individualHarbor/harborProperties/HarborProperties.tsx
TatuArvela: True, I'll remove it

src/common/modal/modal.module.scss
santtul: Could this be like before (the modal adapts to it's content size). What do you think?

src/common/modal/modal.module.scss
santtul: The react modal is rendered as block level element (`div`), so it naturally takes the whole width of the parent container. There is no need to specify `width: 100%` for a block level elements. Do you agree? :)

TatuArvela: Ah, indeed. It also looks to break modals without a defined size, so it makes sense to remove it.

src/common/modal/modal.module.scss
TatuArvela: Maybe that's the best approach after all, I'll do this in a bit

src/common/fileUpload/fileUploadButton/FileUploadButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * The size of the button
+   *
+   * Available options: `'default' | 'small'`
+   */
+  size: 'default' | 'small';
+}
+
+/** Fork of HDS Button. 'button' replaced with 'div' */
+const FileUploadButton: FunctionComponent<FileUploadButtonProps> = ({
```
Nizar-Rahme, line 34: Shoter version :)
```suggestion
const FileUploadButton: React.FC<FileUploadButtonProps> = ({
```

src/domain/individualHarbor/forms/harborForm/utils/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    return {
+      name: imageFile,
+      markedForDeletion: false,
+    };
+  };
+
+  const getMaps = (maps: HARBOR_PROPERTIES['maps']) =>
+    maps.reduce((acc: FileContainer[], map) => {
```
Nizar-Rahme, line 24:
```suggestion
    maps.reduce<FileContainer[]>((acc, map) => {
```

src/domain/individualHarbor/forms/harborForm/utils/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    wwwUrl,
+    imageFile: getImageFile(imageFile),
+    maps: getMaps(maps),
+  };
+};
+
+export const mapValuesToMutation = (harborId: string, values: Harbor) => {
+  const { name, streetAddress, zipCode, municipality, wwwUrl, imageFile, maps } = values;
+
+  const { addMapFiles, removeMapFiles } = (maps || []).reduce(
```
Nizar-Rahme, line 48:
```suggestion
  const { name, streetAddress, zipCode, municipality, wwwUrl, imageFile, maps = [] } = values;

  const { addMapFiles, removeMapFiles } = maps.reduce(
```

src/domain/individualHarbor/forms/harborForm/utils/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    (acc: { addMapFiles: File[]; removeMapFiles: string[] }, map) => {
```
Nizar-Rahme, line 49: Would be good if we extract this inline type `{ addMapFiles: File[]; removeMapFiles: string[] }` as an interface and apply it as a generic to `reduce` function.

src/common/fileUpload/fileUploadButton/FileUploadButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+  className,
+  color,
+  disabled,
+  size,
+  label,
+}: FileUploadButtonProps) => {
+  return (
+    <div className={classNames(styles.button, styles[color], styles[size], disabled ? styles.disabled : '', className)}>
```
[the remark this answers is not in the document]
TatuArvela: Browsers don't consider `button` inside `label` part of the input, and do not proxy the click to the `input`.

Using `button` could be possible, but would require building a custom handler that repeats the event to `input`

src/common/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const renderFileList = () => {
+    if (value === undefined) return undefined;
+
+    const valueList = Array.isArray(value) ? value : [value];
+    return (
+      <List noBullets>
+        {valueList.map((file) => {
+          return (
+            <li key={file.name} className={styles.fileListItem}>
```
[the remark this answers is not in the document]
TatuArvela: This was a great point, I added `uuid` to FileContainers to identify them

src/common/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,205 @@
...
+                ...(file as PersistedFileContainer),
+                markedForDeletion: !file.markedForDeletion,
+              });
+            } else {
+              // New file, remove reference
+              return acc;
+            }
+          } else {
+            return acc.concat(file);
+          }
```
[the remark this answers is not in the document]
TatuArvela: Thanks, added changes

src/common/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,205 @@
...
+      // single
+      return onChange(eventFiles[0]);
+    }
+  };
+
+  const handleDelete = (targetFile: FileContainer) => {
+    // multiple
+    if (Array.isArray(value)) {
+      return onChange(
+        value.reduce((acc: FileContainer[], file: FileContainer) => {
```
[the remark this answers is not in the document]
TatuArvela: Thanks, added

src/common/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,205 @@
...
+        )}
+      </div>
+
+      {helperText && (
+        <div
+          className={classNames({
+            [styles.helperText]: true,
+            [styles.invalid]: invalid,
+          })}
+        >{`${helperText}`}</div>
```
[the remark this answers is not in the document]
TatuArvela: This was to catch possible unknown errors from Yup, but we should probably make a common util for that

src/common/fileUpload/fileUploadButton/fileUploadButton.module.scss
```diff
@@ -0,0 +1,139 @@
...
+  border: var(--hds-ui-color-black-30);
+}
+
+.supplementary:hover {
+  text-decoration: underline;
+}
+
+.small {
+  padding: 0.25em;
+}
```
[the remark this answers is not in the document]
TatuArvela: This is copy-pasted from HDS

src/common/fileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,205 @@
...
+              maxSize: formatBytes(maxSize, i18n.language),
+            })}
+          </Text>
+        )}
+      </div>
+
+      {helperText && (
+        <div
+          className={classNames({
+            [styles.helperText]: true,
```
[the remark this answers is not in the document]
TatuArvela: `Text` doesn't offer identical `line-height` as HDS input helper texts, but otherwise it's fine - changed

src/domain/individualHarbor/forms/harborForm/utils/__tests__/utils.test.ts
```diff
@@ -0,0 +1,170 @@
...
+        id: 'test',
+        municipalityId: 'helsinki',
+        name: 'Pajalahden venesatama (Meripuistotie) / Venesatama',
+        streetAddress: 'Meripuistotie 1a',
+        wwwUrl: 'https://hel.fi',
+        zipCode: '00210',
+      });
+    });
+
+    it('should map empty "imageFile" and "maps" correctly', () => {
```
badrange, line 122: Is the name of this test correct?

TatuArvela: Sharp eye! Thanks, that one's a mistake

src/common/fileUpload/FileUpload.tsx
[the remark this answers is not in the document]
Nizar-Rahme: That works; Or simpler but not an ideal solution is using the elements' indexes since there will be only a handful of files.
```TSX
{valueList.map((file, i) => {
          return (
            <li key={i} className={styles.fileListItem}>
```

src/common/fileUpload/fileUploadButton/FileUploadButton.tsx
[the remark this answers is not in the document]
Nizar-Rahme: Fair enough!

src/common/fileUpload/fileUploadButton/fileUploadButton.module.scss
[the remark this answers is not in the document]
Nizar-Rahme: But using HDS custom variables in our SASS modules will make it hard to trace and update in the future, especially if they introduced some breaking changes as they did since the absence of a CSS variable's value would fail silently. 
That's why we mapped all of their values to our own SASS variables in one partial `_colours.scss`, so when they update their variables we know that we need only to update it in one place.
[These changes](https://github.com/City-of-Helsinki/berth-reservations-admin/pull/188/files#diff-eddb3742e0a6f751810131793ef6a720) in @santtul's PR may explain my point.

PR #186, opened 2020-05-19
VEN-595 | Add header text to all pages

TatuArvela:
## Description :sparkles:

* Add header text to all pages
* Remove duplicate key "forms.berth.delete"

### Closes :no_good_woman:

* [VEN-595](https://helsinkisolutionoffice.atlassian.net/browse/VEN-595) | Add header text to all pages

### Automated tests :gear:️

* Updated snapshots

* Go through all pages and test that the header looks nice and the text is accurate for content


src/domain/pricing/PricingPage.tsx
```diff
@@ -205,36 +206,36 @@ const PricingPage: React.SFC<PricingPageProps> = ({
...
-      </Card>
-      <Card>
-        <CardHeader title={t('pricing.additionalServices.title')} />
-        <CardBody>
-          <Table columns={additionalServicesCols} data={additionalServicesData} theme="basic" />
-        </CardBody>
-      </Card>
+      <PageTitle title={t('pricing.title')} />
+      <div className={styles.grid}>
+        <Card className={styles.fullWidth}>
```
Nizar-Rahme, line 211: Do we need this outer wrapper `div`? 
Can't we do 
```TSX
   <Card className={classNames(styles.fullWidth, styles.grid)}>
```

src/common/pageTitle/pageTitle.module.scss
```diff
@@ -0,0 +1,10 @@
+@import 'colours';
+@import 'spacings';
+
+.pageTitle {
+  color: $gray-500;
+  font-size: 1.25em;
```
Nizar-Rahme, line 6: We have a partial for `font-sizes` that can be used here.
```SCSS
@import 'fonts';

.element {
  font-size: font-size(16);
}
```

src/common/pageTitle/PageTitle.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { FunctionComponent } from 'react';
+
+import styles from './pageTitle.module.scss';
+
+export interface PageTitleProps {
+  title: string;
+}
+
+const PageTitle: FunctionComponent<PageTitleProps> = ({ title }) => {
+  return <h2 className={styles.pageTitle}>{title.toUpperCase()}</h2>;
```
Nizar-Rahme, line 10: Would `Text` component be useful here?
```TSX
<Text as="h2" weight="normalWeight" size="xxl" className={...}>
```

src/domain/pricing/PricingPage.tsx
TatuArvela: That would make every card into a grid of its own, the goal here is to have a grid with the cards in it

src/common/pageTitle/pageTitle.module.scss
TatuArvela: Indeed, but I'll use Text instead so this will not be needed

src/common/pageTitle/PageTitle.tsx
TatuArvela: That'll work. The only aspect that it does not support is the color of the text, so I'll put that into the module style

src/domain/pricing/PricingPage.tsx
Nizar-Rahme: Of course, my bad!

PR #187, opened 2020-05-19
VEN-594 | Add service name to page header

TatuArvela:
## Description :sparkles:

* Add service name to page header

### Closes :no_good_woman:

* [VEN-594](https://helsinkisolutionoffice.atlassian.net/browse/VEN-594) | Add service name to page header

* Static feature, ensure that the title looks correct

### DIFF
--- a/src/domain/page/pageHeader/PageHeaderContainer.tsx
+++ b/src/domain/page/pageHeader/PageHeaderContainer.tsx
@@ -11,6 +11,8 @@ import List from '../../../common/list/List';
 import ListItem from '../../../common/list/ListItem';
 import { useCurrentUser } from '../../auth/hooks';
 import authService from '../../auth/authService';
+import Text from '../../../common/text/Text';
+import styles from './pageHeader.module.scss';
 
 const PageHeaderContainer: React.SFC = () => {
   const { t } = useTranslation();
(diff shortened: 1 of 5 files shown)


src/domain/page/pageHeader/pageHeader.module.scss
```diff
@@ -0,0 +1,15 @@
+@import 'colours';
+@import 'spacings';
+
+.mainLink {
+  display: flex;
+  align-items: center;
+  justify-content: center;
```
santtul, line 7: This style is not needed because the element (`a`) width is based on the content (there is no free space to distribute), so we are fine with the default value `flex-start`, so no additional style needed.

src/domain/page/pageHeader/pageHeader.module.scss
```diff
@@ -0,0 +1,15 @@
...
+  text-decoration: none;
+}
+
+.serviceName {
+  color: $white;
+  display: inline-block;
```
santtul, line 13: Style has no effect, no need to specify as `inline-block` here

src/domain/page/pageHeader/PageHeaderContainer.tsx
```diff
@@ -11,6 +11,8 @@ import List from '../../../common/list/List';
 import ListItem from '../../../common/list/ListItem';
 import { useCurrentUser } from '../../auth/hooks';
 import authService from '../../auth/authService';
+import Text from '../../../common/text/Text';
+import styles from './pageHeader.module.scss';
```
Nizar-Rahme, line 15: I don't recall why this was called a container in the first place or the implementation has been changed somewhere down the line, but now especially that we are applying styles it looks more of a presentational component. So do you think it would be better to rename it? 
`PageHeaderContainer.tsx` => `PageHeader.tsx`

TatuArvela: Makes sense, let's do it

PR #189, opened 2020-05-19
VEN-613 | Replace frontend table sorting with sorted queries from the backend

Nizar-Rahme:
## Description :sparkles:
- Implement a custom hook for supporting backend sorting
- Persist, optionally, the table's sorting state when the data changes
- Refetch the customers & applications data when sortBy state changes

### Closes :no_good_woman:
[VEN-613](https://helsinkisolutionoffice.atlassian.net/browse/VEN-613): Replace frontend table sorting with sorted queries from the backend

### Related :handshake:
- [VEN-586](https://helsinkisolutionoffice.atlassian.net/browse/VEN-586): Support pagination for tables

### Automated tests :gear:️
https://github.com/City-of-Helsinki/berth-reservations-admin/blob/d96590adc2dbd53bc4219abd0c49bed14b158c21/src/common/utils/__tests__/useBackendSorting.test.ts#L5

- Customers page:
  - Sorting the table by name would refetch the query sorted from the backend
  - Sorting other columns is disabled until they're supported in the backend
  - Sorting also resets pagination
- Applications page / Single Application page::
  - Sorting the table by PVM would refetch the query sorted from the backend
  - Sorting other columns is disabled until they're supported in the backend
  - Sorting also resets pagination

## Screenshots :camera_flash:
Screen-Recording-2020-05-20-at-1


src/common/utils/useBackendSorting.ts
```diff
@@ -0,0 +1,35 @@
...
+interface SortBy {
+  id: string;
+  desc?: boolean;
+}
+
+export const useBackendSorting = (onOrderByChange?: Function) => {
+  const [orderBy, setOrderBy] = useState<string>();
+
+  // sortBy: the id of the column to be sorted in the table
+  // orderBy: the value sent to the backend for sorted results
```
TatuArvela, line 14: This naming is at least a little confusing

Nizar-Rahme: For the lack of imagination 😄 
I'll try to come up with better names; suggestions are welcome though!

Nizar-Rahme: Renamed!

src/common/utils/useBackendSorting.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState } from 'react';
+
+import { usePrevious } from './usePrevious';
+
+interface SortedCol {
+  id: string;
+  desc?: boolean;
```
santtul, line 7: Would it make sense to create enum `SortDirection` with values `ASC` and `DESC` ?

src/common/table/Table.tsx
```diff
@@ -91,13 +92,15 @@ const Table = <D extends object>({
   theme = 'primary',
   globalFilter,
   initialState,
+  autoResetSortBy = true,
```
santtul, line 95: Could we simplify by removing this prop because it seems that we could derive the value here:
`autoResetSortBy = onSortedColChange ? false : true` ?

Nizar-Rahme: Yes, you're right, atm it's not serving its purpose, I removed it.

src/common/utils/useBackendSorting.ts
Nizar-Rahme: I defined here the same interface used in `react-table` to avoid transformations.

PR #190, opened 2020-05-20
VEN-624 | Offer page: link the customer's name to the corresponding customer's page

TatuArvela:
## Description :sparkles:

* Hide SSN on Application card
* Hide links on Application card
* Show links on Customer card

### Closes :no_good_woman:

* [VEN-624](https://helsinkisolutionoffice.atlassian.net/browse/VEN-624) | FE - Offer page: link the customer's name to the corresponding customer's page


src/domain/individualApplication/utils.ts
```diff
@@ -20,7 +20,8 @@ export const getCustomerProfile = (profile: CUSTOMER_PROFILE): CustomerProfileCa
     primaryAddress: profile.primaryAddress,
     primaryPhone: profile.primaryPhone?.phone,
     primaryEmail: profile.primaryEmail?.email,
-    ssn: '', // TODO
+    ssn: '-', // TODO
+    showCustomerNameAsLink: true,
```
Nizar-Rahme, line 24: Just realized, if this is always true, can we get rid of it altogether?

TatuArvela: It's not always true, it's false on the Application card

Nizar-Rahme: 👌

PR #191, opened 2020-05-20
VEN-632 | Fix refetchQueries errors after assign/delete berths

Nizar-Rahme:
## Description :sparkles:
- Use the query's name as a string in `refetchQueries`, in order to use the same variables used in the last operation.

## Issues :bug:
See the [conversation in Slack](https://helsinkicity.slack.com/archives/CBA1QHTHN/p1589969426031000)

- Go to the Applications page
- Assign a new berth by following the offer process
- The applications list data should be refetched reflecting the newly assigned berth
- Try to delete an assigned application
- The applications list data should be refetched reflecting the changes


src/domain/applications/queries.ts
```diff
@@ -1,6 +1,9 @@
 import { gql } from 'apollo-boost';
 
 export const BERTH_APPLICATIONS_QUERY = gql`
+  # Note:
+  # The operation name "BERTH_APPLICATIONS" is used as a string in other modules,
+  # for instance, as a value of refetchQueries; If you modify this name, consider changing it in other places too.
```
TatuArvela, line 6: Would there be a way to move the query name in an enum or a constant? That way the dependent refetchQueries would be linked to this query.

Nizar-Rahme: Unfortunately not, it seems that the only way to get the query's name dynamically is using a `getOpertionName` method, which I found it [hidden in Stack Overflow](https://stackoverflow.com/a/55639277/9441362) with no mention in Apollo's documentation whatsoever.
I'll update my PR to use this method.

PR #193, opened 2020-05-26
VEN-625 | Create customer message modal

santtul:
## Description :sparkles:
- Create message modal form and initial layout

### Closes :no_good_woman:
https://helsinkisolutionoffice.atlassian.net/browse/VEN-625
## Testing :alembic:

1. Go to customers
2. Select customer(s)
3. Hit "Lähetä viesti" button
4. Message modal should be shown and validated on submit, fields are mandatory
5. Submit should popup alert which shows the data

Out of scope: we should probably disable the "Lähetä viesti" if no customers selected

### Automated tests :gear:️
Component render test `CustomerListTableTools.test.tsx`


src/domain/customers/tableTools/CustomerMessageForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+      onSubmit={(values) => handleSendMessage?.(values)}
+      validateOnBlur={false}
+      validateOnChange={false}
+      validationSchema={getMessageFormValidationSchema(t)}
+    >
+      {({ values, errors, handleChange, handleSubmit }) => {
+        return (
+          <form onSubmit={handleSubmit} className={styles.form}>
+            <div className={styles.heading}>
+              <Text as="h4">{t('customers.message.header')}</Text>
```
Nizar-Rahme, line 65: This className can be applied directly to `Text` component, no need for this outer `div`.

Though I haven't found this className defined in `customerMessageForm.module.scss`.

src/domain/customers/tableTools/customerMessageForm.module.scss
```diff
@@ -0,0 +1,33 @@
...
+    margin-left: units(0.5);
+  }
+}
+
+.template {
+  width: units(16);
+}
+
+.message {
+  textarea {
```
Nizar-Rahme, line 29: I prefer if we can target this `textarea` via a className instead of its tag name

src/domain/customers/tableTools/customerMessageForm.module.scss
```diff
@@ -0,0 +1,33 @@
+@import 'spacings';
+
+.form {
+  > * {
+    margin: units(1) 0;
+  }
```
Nizar-Rahme, line 6: @ahie defined a mixin that might come in handy here: 
https://github.com/City-of-Helsinki/berth-reservations-admin/blob/d73fc3ac6f15b65080f1e3eb6ec8f9f7d59faa7e/src/assets/styles/_spacings.scss#L44

src/domain/customers/tableTools/customerMessageForm.module.scss
```diff
@@ -0,0 +1,33 @@
...
+  justify-content: space-between;
+}
+
+.formActionButtonsRight {
+  display: flex;
+  justify-content: flex-end;
+
+  > button {
+    margin-left: units(0.5);
+  }
```
Nizar-Rahme, line 21: and here :)

src/domain/customers/tableTools/CustomerMessageForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { useTranslation } from 'react-i18next';
+
+import { MessageFormValues, MessageTemplate } from '../types';
+import Text from '../../../common/text/Text';
+import styles from './customerMessageForm.module.scss';
+import Select from '../../../common/select/Select';
+
+type Props = {
+  closeModal: () => void;
+  handleSendMessage?: (message: MessageFormValues) => void;
```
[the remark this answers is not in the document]
santtul: The reason is that this needs to match the `CustomerListTableToolsProps.handleSendMessage`
and that needs to be optional because the `handleSendMessage` is injected "half way" in the layer where the table is. It's similar to `handleLinkCustomer` in Applications page. This is not that easy to follow, we should maybe consider at some point what to do with this.

src/domain/customers/tableTools/CustomerMessageForm.tsx
santtul: Good catch, the style doesn't exists :D will remove.

src/domain/customers/tableTools/customerMessageForm.module.scss
santtul: We cannot target it with class because of CSS modules (that classname is generated by HDS CSS modules). This is one of the drawbacks with CSS modules :(

src/domain/customers/tableTools/customerMessageForm.module.scss
santtul: Unfortunately I don't have a common classname for the child elements in this case, so I cannot use that mixin.

src/domain/customers/tableTools/customerMessageForm.module.scss
santtul: same here :)

src/domain/customers/tableTools/CustomerListTableTools.tsx
```diff
@@ -0,0 +1,65 @@
...
+          required
+        />
+        <TextInput
+          className={styles.searchInput}
+          id="searchSimilarCustomers"
+          value={searchVal}
+          onChange={(e) => setSearchVal((e.target as HTMLInputElement).value)}
+        />
+      </div>
+      {messageModalOpen && (
```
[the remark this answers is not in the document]
santtul: done.

src/domain/customers/tableTools/customerMessageForm.module.scss
Nizar-Rahme: But the `TextArea` in HDS accepts custom `className`s:

> export declare type TextAreaProps = TextareaHTMLAttributes<HTMLTextAreaElement> & {
    /**
     * Additional class names to apply to the textarea
     */
    className?: string;

src/domain/customers/tableTools/customerMessageForm.module.scss
Nizar-Rahme: We can give these buttons `className`s
https://github.com/City-of-Helsinki/berth-reservations-admin/blob/17ef7750ffcb4b5a04068670e119f9d649a57ed3/src/domain/individualHarbor/individualHarborTableTools/IndividualHarborTableTools.tsx#L20

Nizar-Rahme:
LGTM 👍

src/domain/customers/tableTools/customerMessageForm.module.scss
santtul: Tried to do this with class but HDS TextArea className is applied to the wrapper div only which doesn't work for us :(

PR #195, opened 2020-05-27
File upload refactoring

TatuArvela:
## Description :sparkles:

* Refactored FileUpload into separate FileList and FileUpload components
* Use HDS Button in FileUpload
* Adjusted icons stories typing

### Closes :no_good_woman:

* [VEN-575](https://helsinkisolutionoffice.atlassian.net/browse/VEN-575): Single harbor page: edit harbor properties

### Automated tests :gear:️

* Tests have been updated, and should have 100% coverage

* Go to an individual harbor's page and edit the file properties
  * File upload should create file entities correctly, but the actual data is not correctly saved into backend (needs further investigation)
* The components can be tested in Storybook as well (FileList, FileUpload)


src/common/fileList/FileList.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    // single
+    return props.onChange({
+      ...targetFile,
+      markedForDeletion: !targetFile.markedForDeletion,
+    });
+  };
+
+  if (value === undefined) return null;
```
santtul, line 59: This "guard if clause" could be moved right at the beginning of the function.

src/common/fileList/fileList.module.scss
```diff
@@ -0,0 +1,22 @@
...
+
+.fileListItem {
+  margin: units(0.25) 0;
+  display: flex;
+  align-items: center;
+}
+
+.fileName {
+  flex-grow: 1;
+  line-height: font-size(24);
```
santtul, line 17: We are not setting font-size here so could we just write `24px` ?

src/domain/individualHarbor/forms/harborForm/utils/utils.ts
```diff
@@ -41,39 +38,31 @@ export const getHarbor = (harborData: HARBOR_FORM | undefined): Harbor | undefin
...
+    addedImageFile,
+    existingMaps = [],
+    addedMaps = [],
+  } = values;
 
-      return acc;
-    },
-    { addMapFiles: [], removeMapFiles: [] }
-  );
+  const removeMapFiles = existingMaps.reduce<string[]>((acc, map) => {
```
santtul, line 60: This can be written with good old `filter` like this:
```suggestion
  const removeMapFiles = existingMaps.filter((map) => map.markedForDeletion).map((map) => map.id);
```

src/common/fileList/fileList.module.scss
TatuArvela: We're trying to avoid setting sizes outside the standard spacings, but good catch. `units(1.5)` gives the same result and makes more sense than `font-size(24)`

src/domain/individualHarbor/forms/harborForm/utils/utils.ts
TatuArvela: That's true, good suggestion!

PR #200, opened 2020-05-29
VEN-654 | Implement generic error handling

ahie:
## Description :sparkles:

Adds generic error handling and fallback route to the app.

### Closes :no_good_woman:

[VEN-654](https://helsinkisolutionoffice.atlassian.net/browse/VEN-654)

### Automated tests :gear:️
`src/domain/errorPage/__tests__/ErrorPage.test.tsx`
`src/common/errorBoundary/__tests__/ErrorBoundary.test.tsx`
`src/common/toast/__tests__/HDSToastContainer.test.tsx`

- An error page should be visible at route `/error`
  - The ErrorBoundary should redirect to this page on any uncaught errors during component rendering.
  - Can be tested manually by building the project with a component that throws intentionally.
- Network disconnects should display a warning toast.
- GraphQL errors should display an error toast. Can be tested manually by for example attempting to delete a berth with an active lease.


src/domain/errorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  return (
+    <div className={styles.container}>
+      <div className={styles.innerContainer}>
+        <Icon shape="IconExclamationCircle" size="large" />
+        <Text as="h1" className={styles.heading}>
+          {t('errorPage.heading')}
+        </Text>
+        <Text>{t('errorPage.text')}</Text>
+        <InternalLink to={'/'} className={styles.link}>
```
santtul, line 20: Can you make this ExternalLink to force the user to reload the SPA after the error.

ahie: Done 👍

ahie:
There are probably places in the code that could be redirected to this error page still left in the code. Please add them as comments here. The error boundary added here should be enough to catch uncaught exceptions from our functional components.

src/domain/app/App.tsx
```diff
@@ -26,14 +28,17 @@ const App: React.FC = () => {
...
+              <Route exact path="/error" component={ErrorPage} />
+              <ErrorBoundary errorComponent={<Redirect to="/error" />}>
+                <PrivateRoute exact path="/harbors/:id" component={IndividualHarborPage} />
+                <PrivateRoute exact path="/harbors" component={HarborsPage} />
+                <PrivateRoute exact path="/customers/:id" component={IndividualCustomerPage} />
+                <PrivateRoute exact path="/customers" component={CustomersPage} />
+                <PrivateRoute exact path="/applications/:id" component={IndividualApplicationPage} />
+                <PrivateRoute exact path="/applications" component={ApplicationsPage} />
+                <PrivateRoute exact path="/offer/:applicationId" component={OfferPage} />
+                <PrivateRoute exact path="/pricing" component={PricingPage} />
```
Nizar-Rahme, line 40: These `PrivateRoute`s should be wrapped directly by a `Switch` component; otherwise, the router would try to render all of them.
You can test it, for example, by adding a `console.log('private route')` statement inside the `PrivateRoute` component, it will be called eight times in this case, while if we wrap them with a `Switch` component, it will be called only once.

ahie: Fixed.

src/common/errorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+// Error boundaries have to be implemented as class components.
+class ErrorBoundary extends React.Component<ErrorBoundaryProps, ErrorBoundaryState> {
+  constructor(props: { errorComponent: React.ReactNode }) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  static getDerivedStateFromError() {
+    return { hasError: true };
```
Nizar-Rahme, line 19: Should we send caught errors to Sentry? 🤔

ahie: This shouldn't interfere with Sentry (the error still gets caught by Sentry's default handler), just prevents react from tearing down the page. I'll test that this is actually the case.

src/domain/errorPage/errorPage.module.scss
```diff
@@ -0,0 +1,25 @@
+@import 'spacings';
+
+.container {
+  width: 100%;
```
santtul, line 4: Style has no effect

src/domain/errorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
```
santtul, line 1: The left menu should be not visible in the error page if you check the layout specified in Abstract.

src/common/errorBoundary/ErrorBoundary.tsx
ahie: The errors caught by the boundary did infact need to be explicitly sent to Sentry. Fixed.

src/domain/errorPage/errorPage.module.scss
ahie: Removed.

src/domain/errorPage/ErrorPage.tsx
ahie: Fixed.

src/common/errorBoundary/ErrorBoundary.tsx
Nizar-Rahme: 👍

src/app/App.tsx
```diff
@@ -15,27 +15,35 @@ import ApplicationListPage from '../features/applicationList/ApplicationListPage
...
+                <Switch>
+                  <PrivateRoute exact path="/harbors/:id" component={HarborViewPage} />
+                  <PrivateRoute exact path="/harbors" component={HarborListPage} />
+                  <PrivateRoute exact path="/customers/:id" component={CustomerViewPage} />
+                  <PrivateRoute exact path="/customers" component={CustomerListPage} />
+                  <PrivateRoute exact path="/applications/:id" component={ApplicationViewPage} />
+                  <PrivateRoute exact path="/applications" component={ApplicationListPage} />
+                  <PrivateRoute exact path="/offer/:applicationId" component={OfferPage} />
+                  <PrivateRoute exact path="/pricing" component={PricingPage} />
+                  <PrivateRoute exact path="/error" component={ErrorPage} />
```
santtul, line 42: Is this useless duplicate route?

ahie: Good catch. Removed.

PR #212, opened 2020-06-09
VEN-683 | Back button issues

santtul:
## Description :sparkles:
Fixes following issues
- Need to click back three times when opening application that has not been attached to customer
- Need to click back several times after you have used customer search field
- Forces to first page regardless of the url page param on customer list

### Closes :no_good_woman:
https://helsinkisolutionoffice.atlassian.net/browse/VEN-683

## Testing :alembic:
- Open application that has not been attached to customer, click back, it should take you to applications list
- Go to customer list, search with search box, click back, it should take you to previous screen
- Go to customer list, browse to third page, hit reload in browser, you should stay on third page

### Automated tests :gear:️
Added test case to `usePagination.test.tsx`

### DIFF
--- a/src/common/utils/usePagination.ts
+++ b/src/common/utils/usePagination.ts
@@ -1,13 +1,12 @@
 import { useHistory } from 'react-router-dom';
 import { useCallback } from 'react';
+import { History } from 'history';
 
 const PAGE_SIZE = 10;
 
 export const usePagination = (pageSize = PAGE_SIZE) => {
   const history = useHistory();
-
-  const urlSearchParams = new URLSearchParams(history.location.search);
-  const pageParam = Number(urlSearchParams.get('page') ?? 1);
+  const pageParam = getCurrentPageParam(history);
   const pageIndex = !Number.isNaN(pageParam) ? pageParam - 1 : 0;
   const cursorIndex = pageIndex * pageSize;
   const cursor = cursorIndex > 0 ? btoa(`arrayconnection:${cursorIndex - 1}`) : undefined;
@@ -15,7 +14,17 @@ export const usePagination = (pageSize = PAGE_SIZE) => {
   const getPageCount = (connectionsCount: number | null | undefined) =>
     connectionsCount ? Math.ceil(connectionsCount / pageSize) : 1;
 
-  const goToPage = useCallback((pageIndex: number) => history.push({ search: `?page=${pageIndex + 1}` }), [history]);
+  const goToPage = useCallback(
+    (pageIndex: number) => {
+      const newPageNumber = pageIndex + 1;
+      const currentPageNumber = getCurrentPageParam(history);
+
+      if (currentPageNumber !== newPageNumber) {
+        history.push({ search: `?page=${newPageNumber}` });
+      }
+    },
+    [history]
+  );
 
   return {
     cursor,
@@ -24,4 +33,9 @@ export const usePagination = (pageSize = PAGE_SIZE) => {
     getPageCount,
     goToPage,
   };
+};
+
+const getCurrentPageParam = (history: History<History.LocationState>): number => {
+  const urlSearchParams = new URLSearchParams(history.location.search);
+  return Number(urlSearchParams.get('page') ?? 1);
 };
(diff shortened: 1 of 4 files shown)


src/common/utils/usePagination.ts
```diff
@@ -1,21 +1,30 @@
...
     connectionsCount ? Math.ceil(connectionsCount / pageSize) : 1;
 
-  const goToPage = useCallback((pageIndex: number) => history.push({ search: `?page=${pageIndex + 1}` }), [history]);
+  const goToPage = useCallback(
+    (pageIndex: number) => {
+      const newPageNumber = pageIndex + 1;
+      const currentPageNumber = getCurrentPageParam(history);
+
+      if (currentPageNumber !== newPageNumber) {
+        history.push({ search: `?page=${newPageNumber}` });
```
ahie, line 23: Not directly related to the issue, but would it make sense to use `history.replace` here instead? Would make navigating between different views easier, but disable navigating table pages with the browser's history. Alternatively this could be toggled with an optional prop to the hook.

santtul: Good point. In many web apps back leads you to previous paginated page (push is used), but it may not be the best for all apps, it's hard to say without user feedback :)

src/common/utils/usePagination.ts
```diff
@@ -1,21 +1,30 @@
...
+      }
+    },
+    [history]
```
Nizar-Rahme, line 26: According to [useCallback documentation](https://reactjs.org/docs/hooks-reference.html#usecallback):

>  every value referenced inside the callback should also appear in the dependencies array.

src/common/utils/usePagination.ts
```diff
@@ -1,21 +1,30 @@
...
   const cursor = cursorIndex > 0 ? btoa(`arrayconnection:${cursorIndex - 1}`) : undefined;
 
   const getPageCount = (connectionsCount: number | null | undefined) =>
     connectionsCount ? Math.ceil(connectionsCount / pageSize) : 1;
 
-  const goToPage = useCallback((pageIndex: number) => history.push({ search: `?page=${pageIndex + 1}` }), [history]);
+  const goToPage = useCallback(
+    (pageIndex: number) => {
+      const newPageNumber = pageIndex + 1;
+      const currentPageNumber = getCurrentPageParam(history);
```
Nizar-Rahme, line 20: Why do we need to extract the logic to `getCurrentPageParam`? why don't we just access `pageParam` here?

src/features/customerList/CustomerListPageContainer.tsx
```diff
@@ -45,9 +45,16 @@ const CustomerListPageContainer: React.FC = () => {
...
+  const isInitialMount = useRef(true);
   useEffect(() => {
-    // Go to the first page when search values change.
-    goToPage(0);
+    // Prevent hook running on initial mount because it would force to first page on landing with direct url
+    // regardless of the page param
+    if (isInitialMount.current) {
+      isInitialMount.current = false;
+    } else {
+      // Go to the first page when search values
```
Nizar-Rahme, line 55:
```suggestion
      // Go to the first page when search values change.
```

src/common/utils/usePagination.ts
santtul: The `pageParam` that is visible as closure is not updated after the component is initialized, so we need to get it from URL.

src/common/utils/usePagination.ts
santtul: I think that all the dependencies are listed in array, `history` is the only one. Others are just local consts.

Nizar-Rahme: How about `getCurrentPageParam`?

santtul: That one cannot change and it's a function, so it would make no difference. Linter is happy also: this is in compliance with `react-hooks/exhaustive-deps` rule.

src/common/utils/usePagination.ts
Nizar-Rahme: Makes sense.

PR #213, opened 2020-06-10
Remove React.FC

santtul:
## Description :sparkles:
As agreed with the team:

- Remove `React.FC`
- Remove `React.SFC`
- Remove `React.FunctionComponent`

## Testing :alembic:
It should compile :stuck_out_tongue_winking_eye:


santtul:
Good catch! Fixed.

PR #220, opened 2020-06-17
VEN-636 | Use HDS Dropdown component

TatuArvela:
## Description :sparkles:

* Replace the custom functionality of Select with a wrapper for HDS's Dropdown component
* Extra: Use the shared `periodType` mapping in pricing forms (used with dropdowns)

### Closes :no_good_woman:

* [VEN-636](https://helsinkisolutionoffice.atlassian.net/browse/VEN-636): Use HDS drop down

### Automated tests :gear:️

Automated tests and snapshots were updated.

Check that each form works as expected.

Dropdown components are used in:
  * Customer list
  * Berth editing
  * Pricing editing (all types)


src/common/select/Select.tsx
```diff
@@ -1,47 +1,45 @@
-import React from 'react';
-import classNames from 'classnames';
-
-import styles from './select.module.scss';
+import React, { ChangeEvent } from 'react';
+import { Dropdown } from 'hds-react';
 
 interface Option {
   label: string;
   value: string | number;
```
ahie, line 6: > value: string | number;

API suggestion: could this value be made generic? This would ofcourse also need a converter from the generic type to string, which could simply default to `() => String(value)` if none is provided.

src/common/select/Select.tsx
```diff
@@ -1,47 +1,69 @@
 import React from 'react';
-import classNames from 'classnames';
+import { Dropdown } from 'hds-react';
 
-import styles from './select.module.scss';
+type OptionValue = string | number;
```
ahie, line 4: > type OptionValue = string | number;

Optional: This could be any object if we choose to do deep comparisons. Though restricting it to `string` or `number` covers our current use cases and has a cleaner implementation.

TatuArvela: The current typing allows specifying enums, so I think this covers our needs well enough. Deep comparison of objects could open some possibilities, but I'm not sure if it has any concrete uses for us right now.  Select options can be mapped from more complex objects

PR #227, opened 2020-06-25
VEN-607 | Use coat-of-arms-blue for Checkbox and fix its line height 

TatuArvela:
## Description :sparkles:

* Use coat-of-arms-blue for Checkbox and fix its line height
* Update all Sass colour variables to match HDS naming

### Closes :no_good_woman:

* [VEN-607](https://helsinkisolutionoffice.atlassian.net/browse/VEN-607): Replace Checkbox with the one from HDS

### Automated tests :gear:️

* Updated changed snapshots (new className added)

* Verify that the appearance of checkboxes is now correct


src/assets/styles/_colours.scss
```diff
@@ -1,27 +1,21 @@
...
+$black-5: var(--color-black-5);
+$black-10: var(--color-black-10);
+$black-20: var(--color-black-20);
+$black-30: var(--color-black-30);
+$black-40: var(--color-black-40);
+$black-50: var(--color-black-50);
+$black-60: var(--color-black-60);
+$black-70: var(--color-black-70);
+$black-80: var(--color-black-80);
+$black-90: var(--color-black-90);
```
Nizar-Rahme, line 18: IMO, having the same scale for colors is better for the sake of consistency, even if it is different from what HDS uses, since HDS doesn't look like using any convention for naming colors, and we're mapping their values anyway.

Unlike having one scale for all colors (e.g. x-100, x-200, ..., x-900), with this change, we would end up with inconsistent names ($blue-light-50, $blue, $blue-dark-50, $black-20, $black-30, $orange-200, $orange-500)
Therefore, I'm still convinced that our current naming convention is better than HDS's.

Nizar-Rahme: What do you and @ahie think?

TatuArvela: The colour system is available here: https://hds.hel.fi/design-tokens/colour

I agree that there's some inconsistency in HDS's colours, which makes them harder to memorize. Some brand colours have 5 variants, some have 4 variants, and Coat of Arms Silver has only 1 variant. UI colours all have 3 variants. Grayscale has 12 variants. In general though, every colour (excluding Silver) has at least a light and a dark variant, while the naming between the systems is a little different.

By adopting HDS naming in the codebase, we could discuss colours values with the same terminology. It's worth pointing out that we use a fairly limited set of the HDS colours (Coat of Arms Blue, UI colours, grayscale).

For the grayscale, the difference is only removing a trailing '0'. Something we could do is define some internal naming for the brand colour we use, like `brand`, `brand-dark`, `brand-light`, `brand-lighter`, that way the UI colours and brand colour would follow the same pattern (excluding `-lighter`).

`orange` is not in the HDS, and should probably be replaced with `alert`, `summer` or `metro`. It's only used as the colour for the application status `NO_SUITABLE_BERTHS_NOTIFIED`.

Most of the defined colours are only used in Chip, and some were not used at all (which I removed).

ahie: IMO we should use the variables provided by HDS, or map them directly like in this PR (dropping the `var(—color-)` part makes usage more concise), as they are (or should be?) generated from the design tokens. Also avoiding custom colors that are not defined in HDS whenever possible should be a goal as well? Though this is a larger discussion and should involve Ville Hirvonen. I think this refactoring is good enough for now, as any further changes would be out of scope for this PR.

Nizar-Rahme: Fair enough! 
Could you @TatuArvela please rename `$orange-200` and `$orange-500` to `$orange-20` and `$orange-50` respectively?

TatuArvela: @Nizar-Rahme renamed 👍

PR #231, opened 2020-06-29
VEN-622, VEN-731 | Fetch order information from BE for BillsCard and BillingHistoryCard

ahie:
## Description :sparkles:

After this patch billing information is fetched from the backend in the customer page.

### Closes :no_good_woman:

[VEN-622](https://helsinkisolutionoffice.atlassian.net/browse/VEN-622)
[VEN-731](https://helsinkisolutionoffice.atlassian.net/browse/VEN-731)

### Automated tests :gear:️

- Updated `BillsCard.test.jsx`.
- Added `BillingHistoryCard.test.tsx`

- Open `/customers/UHJvZmlsZU5vZGU6MmI1YzViMmItM2FjMi00MTAzLWI2ODAtZjIxZWJmODBlNDMy`
- Open bill information should be displayed as provided by the backend.
- Billing history should be displayed as provided by the backend.

The "Open bill" button is still a placeholder, will need to be implemented once the page/modal it leads to is done.


src/features/customerView/CustomerViewContainer.tsx
```diff
@@ -41,6 +41,7 @@ const CustomerViewContainer = () => {
   const leases = getLeases(data.profile);
   const boats = getBoats(data.profile);
   const applications = getApplications(data.profile, data.boatTypes || []);
+  const openBills: BerthBill[] = getOpenBills(data.profile);
```
Nizar-Rahme, line 44: Doesn't TS infer `BerthBill[]` type here?

src/features/customerView/queries.ts
```diff
@@ -76,6 +76,25 @@ export const INDIVIDUAL_CUSTOMER_QUERY = gql`
...
+              totalPrice
+              totalTaxPercentage
+              price
+              taxPercentage
+              status
+              orderLines {
+                edges {
+                  node {
+                    product {
+                      service
```
Nizar-Rahme, line 90:
```suggestion
                edges {
                  node {
                    id
                    product {
                      id
                      service
```

src/features/customerView/queries.ts
```diff
@@ -76,6 +76,25 @@ export const INDIVIDUAL_CUSTOMER_QUERY = gql`
                 }
               }
             }
+            order {
```
Nizar-Rahme, line 79: Apollo sometimes causes weird issues with `id`s are not queried. 
```suggestion
            order {
```

src/features/customerView/utils.ts
```diff
@@ -140,3 +141,78 @@ export const getApplications = (profile: CUSTOMER_PROFILE, boatTypes: BOAT_TYPES
...
+    harborName: string;
+  };
+};
+
+export const getOpenBills = (profile: CUSTOMER_PROFILE): BerthBill[] => {
+  return (
+    profile.berthLeases?.edges
+      .map((edge) => edge?.node)
+      .filter((node) => node?.order?.status === OrderStatus.WAITING)
+      .reduce<BerthBill[]>((acc, berthLeaseNode) => {
```
Nizar-Rahme, line 177: Since we're calling reduce here, I think all of these three operations can be achieved in `reduce`'s callback

src/features/customerView/CustomerViewContainer.tsx
ahie: Yes, removed. Was added there just to test an empty list and forgot to remove it 😞

src/features/customerView/queries.ts
ahie: I don't think this is relevant anymore, as the query now fetches `orders` instead of an `order` of a lease.

src/features/customerView/queries.ts
ahie: Will add `id`s.

src/features/customerView/utils.ts
ahie: The filter has now been removed and the function just gets all bills.

PR #233, opened 2020-06-30
Fix appearance of HDS Select, Checkbox and RadioButton outside local env

TatuArvela:
## Description :sparkles:

* Increase specifity of selectors of 'Select' and 'Checkbox'
* Use coat-of-arms-blue for RadioButton 

* Verify correct appearance of Select (visible on pricing page, for harbor services)
* Verify correct appearance of Checkbox (visible on customer list page)
* Verify correct appearance of RadioButton (visible on offer page, when selecting customer profile)

### DIFF
--- a/src/common/checkbox/checkbox.module.scss
+++ b/src/common/checkbox/checkbox.module.scss
@@ -1,6 +1,6 @@
 @import 'colours';
 
-.checkbox {
+div.checkbox {
   line-height: var(--lineheight-l);
   --background-selected: #{$blue};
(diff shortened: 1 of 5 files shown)


TatuArvela:
> LGTM!
> Though I haven't really understood the root cause of the issue 😅

In production(esque) environments, the HDS styles are loaded after our styles. This gives them the minor upper hand in the priority and causes their styles to skip our overrides. Increasing the specificity (`div.class` instead of just `.class`) of our selectors fixes this issue.

PR #247, opened 2020-07-30
VEN-769 | Add ability to create boats for customers

ahie:
## Description :sparkles:

Adds `BoatCreateForm` which allows for the creation of new boats.

### Closes :no_good_woman:

[VEN-769](https://helsinkisolutionoffice.atlassian.net/browse/VEN-769)

- Open a customer's page
- An "Add boat" button should be present in the customer's boats card
- It should be possible to create a new boat by submitting the form with correctly validated values

"Fixme" changes in `src/features/customerView/utils.ts` and `src/features/customerView/queries.ts` are addressed in a separate PR.


src/locales/fi.json
```diff
@@ -486,6 +486,7 @@
       "title": "ASIAKKAAN VENEET",
       "boatInfo": "VENEEN TIEDOT",
       "editBoat": "Muokkaa",
+      "createBoat": "Lisää uusi",
```
igordavydsson, line 489: it's a common button present in other sections as well, I would either put it in `"common"` or in `"customerView"`

ahie: Will change both `editBoat` and `createBoat` to one from common.

ahie: "Lisää uusi" did not actually exist in the translations before, only "Lisää" was under common. In any case, both "Muokkaa" and "Lisää uusi" are now resolved from common.

src/features/customerView/utils.ts
```diff
@@ -169,10 +169,11 @@ export const getBills = (profile: CUSTOMER_PROFILE): (BerthBill | WinterStorageB
           ...acc,
           {
             status: orderNode.status,
+            // FIXME: resolve berthInformation correctly from new schema.
```
ahie, line 172: Addressed in #249

TatuArvela:
Hmm, creating a boat failed with `Unhandled Rejection (Error): GraphQL error: CustomerProfile matching query does not exist.`

ahie:
> Hmm, creating a boat failed with `Unhandled Rejection (Error): GraphQL error: CustomerProfile matching query does not exist.`

This is probably related to how the backend currently deals with profiles, related discussion here: https://helsinkicity.slack.com/archives/CBA1QHTHN/p1595420856013000

In short, if a customer does not have a profile object in the db, the mutation will fail. The form should work with for example "Arja Häkkinen".

PR #251, opened 2020-08-05
VEN-761 | Use customerGroup field to determine customer group

santtul:
## Description :sparkles:
Use customerGroup field to determine customer group.

### Closes :no_good_woman:
https://helsinkisolutionoffice.atlassian.net/browse/VEN-761

### Automated tests :gear:️
- Unit tests in: `translations.test.ts`
- Several component render tests

- Select "Asiakkaat" from the left menu
- You should see that "Asiakasryhmä" column is populated
- Pro tip: go to the last page and you should see customers with different groups there

### DIFF
--- a/src/@types/__generated__/globalTypes.ts
+++ b/src/@types/__generated__/globalTypes.ts
@@ -90,13 +90,6 @@ export enum OrderStatus {
   WAITING = "WAITING",
 }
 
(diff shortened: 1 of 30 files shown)


santtul:
Good point. Should be removed now.

PR #252, opened 2020-08-06
VEN-753 | Improve test coverage of application list and application view features

TatuArvela:
## Description :sparkles:

* Improve test coverage of application list and application view features
* Split SimilarCustomersTable from ApplicationView  
* Improve test coverage of common/utils
* Refactor features to presentational/container style (applicationList, applicationView, offer, winterStorage)

### Closes :no_good_woman:

* [VEN-753](https://helsinkisolutionoffice.atlassian.net/browse/VEN-753): Improve test coverage of application list and application view features

### Automated tests :gear:️

* **Yes**
* Added automated tests for application list and application view presentational components, containers, subcomponents and utils
  * Test coverage for these features is not yet perfect, but is considerably better than before

* Check that the refactored features work (applicationList, applicationView, offer, winterStorage)


src/features/applicationList/__tests__/utils.test.ts
```diff
@@ -0,0 +1,106 @@
...
+});
+
+describe('utils', () => {
+  describe('getBerthApplicationData', () => {
+    it('should return berth application data', () => {
+      expect(getBerthApplicationData(mockData)).toMatchSnapshot();
+    });
+
+    it('should use defaults for nullable fields', () => {
+      const testEdge: BerthApplicationNodeEdge = {
```
santtul, line 28: Could we move this to `__fixtures__/mockData.ts`?

src/features/applicationView/ApplicationView.tsx
```diff
@@ -1,27 +1,26 @@
...
 import Chip from '../../common/chip/Chip';
 import { APPLICATION_STATUS } from '../../common/utils/consonants';
 import CustomerProfileCard, { CustomerProfileCardProps } from '../../common/customerProfileCard/CustomerProfileCard';
 import OfferCard, { OfferCardProps } from './offerCard/OfferCard';
 import { OrganizationType } from '../../@types/__generated__/globalTypes';
-import Pagination, { PaginationProps } from '../../common/pagination/Pagination';
 import PageTitle from '../../common/pageTitle/PageTitle';
 import PageContent from '../../common/pageContent/PageContent';
 import ActionHistoryCard from '../../common/actionHistoryCard/ActionHistoryCard';
+import SimilarCustomersTableContainer, {
```
santtul, line 21: Should we move the similar customers stuff right under features folder (`features/simarCustomers`). It starts to look like a feature with own queries and mutations. Should the feature be called something like "linkCustomerTable" or "linkCustomer", what do you think?

src/features/applicationView/similarCustomersTable/__tests__/utils.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        profiles: {
+          __typename: 'ProfileNodeConnection',
+          count: 0,
+          edges: [],
+        },
+      };
+      expect(getFilteredCustomersData(testData)).toEqual([]);
+    });
+
+    it('should remove the edges with a value of null', () => {
```
santtul, line 22: This can actually never happen, the basic problem is that the GraphQL contract is wrong from this point of view. So it might be waste to create a test for it.

src/features/applicationView/similarCustomersTable/__tests__/utils.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        profiles: {
+          __typename: 'ProfileNodeConnection',
+          count: 0,
+          edges: [null],
+        },
+      };
+      expect(getFilteredCustomersData(testData)).toEqual(expect.not.arrayContaining([null]));
+    });
+
+    it('should remove the nodes with a value of null', () => {
```
santtul, line 33: Same here.

src/features/applicationView/similarCustomersTable/__tests__/utils.test.ts
santtul: And this comment applies to all places we deal with edges and nodes.

src/features/applicationList/__tests__/utils.test.ts
TatuArvela: Yes. I included it into the set of mock data and changed some field values

src/features/applicationView/similarCustomersTable/__tests__/utils.test.ts
TatuArvela: If this can indeed _never_ happen, should we instruct TypeScript with `as`? For example `(edge as PROFILE_EDGE).node as PROFILE_NODE`.

Leaving the `if (!edge?.node) return acc;` would leave the untested branch noise into the test coverage table.

TatuArvela: Another impossibility I found is `t` function returning `undefined` in Headers of `SimilarCustomersTable`, which I handled with `as`. It would require an invalid localization config to occur, and the app would blow up elsewhere before reaching that point

src/features/applicationView/ApplicationView.tsx
TatuArvela: Makes sense, I did also think about it. How would `linkApplicationToCustomer` sound?

`offerCard/editForm` also has its own queries and mutations, would it make sense to move that one as well?

santtul: The name you suggested sounds great 👍 
and I think moving "offerCard/editForm" would make sense, if you want to do it in this PR.

src/features/applicationView/similarCustomersTable/__tests__/utils.test.ts
santtul: Great idea, let's try the `as`

src/features/applicationView/ApplicationView.tsx
TatuArvela: The forms could be moved later, there are several other "fully featured" forms under other features as well

TatuArvela:
@santtul Rebased now, it was rather painful 😅

PR #265, opened 2020-08-24
Fix browser tests

santtul:
## Description :sparkles:
Make WS and harbors browser test a bit more lenient of data


Nizar-Rahme (approved):
:shipit:

PR #267, opened 2020-09-04
Add temporary fix for fetching customer WS leases

santtul:
## Description :sparkles:
Because of [VEN-813](https://helsinkisolutionoffice.atlassian.net/browse/VEN-813) we should fetch ws lease `section` instead of `area`. As temporary fix remove the `area` reference to get the query working. We need to fix the whole lease names resolving because the implementation did not take marked ws leases in account (only resolved the unmarked names via area). Will create a separate bug issue from that.

### Related :handshake:
[VEN-813](https://helsinkisolutionoffice.atlassian.net/browse/VEN-813)
- land to `http://localhost:3000/applications/QmVydGhBcHBsaWNhdGlvbk5vZGU6NjY=`
- you should be able to search customer to attach to


TatuArvela (approved):
Looks ok 👍

PR #268, opened 2020-09-08
Use temporarily direct BE URL

igordavydsson:
## Description :sparkles:

As the proper URL for the BE is behind a nginx proxy managed
by Kanslia, which is pending necessary config updates, we
could temporarily use the original URL that is provided directly
by Kuva's infra.

Needed to get further testing going by the end users in prod.


santtul (approved):
👍

PR #271, opened 2020-09-10
VEN-834 | Fix create customer button on the single application page

ahie:
## Description :sparkles:

After sending createProfile mutation the button should also call createBerthServicesProfile, which will also create a profile in the venepaikka backend.

### Closes :no_good_woman:

[VEN-834](https://helsinkisolutionoffice.atlassian.net/browse/VEN-834)

### Automated tests :gear:️

`src/common/hooks/__tests__/useCreateNewCustomer.test.tsx `

- Open a single application (either winter storage or harbor) that does not have a connected customer
- Click the "Add new customer" button
- (Network inspector should show a request made to `createProfile` followed by a request to `createBerthServicesProfile` with the input id corresponding to the returned id value from `createProfile`)
- It should be possible to connect the application to the newly created user

## Notes

`OrderStatus` enum has been extended in the backend, which also required changes to a few files after regeneration of types.

### DIFF
--- a/src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
+++ b/src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
@@ -14,11 +14,19 @@ import { usePrevious } from '../../common/utils/usePrevious';
 import { usePagination } from '../../common/utils/usePagination';
 import { useBackendSorting } from '../../common/utils/useBackendSorting';
 import LinkApplicationToCustomer from './LinkApplicationToCustomer';
-import { CREATE_NEW_PROFILE_MUTATION } from '../../common/mutations/createProfile';
+import {
+  CREATE_BERTH_SERVICE_PROFILE_MUTATION,
+  CREATE_NEW_PROFILE_MUTATION,
+} from '../../common/mutations/createProfile';
 import {
   CREATE_NEW_PROFILE,
   CREATE_NEW_PROFILEVariables as CREATE_NEW_PROFILE_VARS,
 } from '../../common/mutations/__generated__/CREATE_NEW_PROFILE';
+import {
+  CREATE_BERTH_SERVICE_PROFILE,
+  CREATE_BERTH_SERVICE_PROFILEVariables as CREATE_BERTH_SERVICE_PROFILE_VARS,
+} from '../../common/mutations/__generated__/CREATE_BERTH_SERVICE_PROFILE';
+import { AddressType, OrganizationType } from '../../@types/__generated__/globalTypes';
 
 export interface LinkApplicationToCustomerContainerProps {
   application: {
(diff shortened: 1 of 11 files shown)


igordavydsson:
one thing to note: some of the application are from company customers and the create customer button should handle that case 😅  namely, address fields (address, zip code and municipality) should not go into `createProfile` mutation but should be part of the `organization` input dict going to `createBerthServicesProfile`. also, company name and business id should be part of the organization input in that case. it wasn't mentioned on the ticket, so, if it feels like a totally different task, we can create a new ticket for this.

ahie:
So, to clarify if `application.businessId !== ''` the mutation should use the following input:

```
input: {
  id: data.createProfile.profile.id
  organization: {
    organizationType: OrganizationType.Company,
    businessId: application.businessId,
    name: application.companyName,
    address: <the one currently used in createProfile?>,
    postalCode: <the one currently used in createProfile?>,
    city: <the one currently used in createProfile?>,
  }
}
```

and remove the address info from `createProfile`, otherwise keep it as is currently?

igordavydsson:
yes, exactly like that 👍

ahie:
Done.

igordavydsson:
nice, in terms of business logic should be good now 👍   frontend devs should check the code 😅

src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
```diff
@@ -38,6 +48,7 @@ const LinkApplicationToCustomerContainer = ({
   application,
   handleLinkCustomer,
 }: LinkApplicationToCustomerContainerProps) => {
+  console.log(application);
```
Nizar-Rahme, line 51: Did you leave it here on purpose?

src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
```diff
@@ -90,19 +113,66 @@ const LinkApplicationToCustomerContainer = ({
...
+                organizationType: OrganizationType.COMPANY,
+                address,
+                postalCode: zipCode,
+                city: municipality,
+              },
+            },
+          },
+        });
+      });
+    }
```
Nizar-Rahme, line 175: One optional note, the same logic might be used in other pages for creating a customer, so we might consider extracting this function now or later to some external module.

Nizar-Rahme (changes requested):

src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
ahie: Nope. Whoops.

Nizar-Rahme (approved):
LGTM!

src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
ahie: Done.

Nizar-Rahme (approved):

PR #282, opened 2020-10-02
VEN-773 | Customer list: Add new customer

TatuArvela:
## Description :sparkles:

* Customer list: Add new customer
* Customer list: Show name and municipality from organization data

### Closes :no_good_woman:

* [VEN-773](https://helsinkisolutionoffice.atlassian.net/browse/VEN-773): FE - Customer list: Add new customer 

* Navigate to customer list
* Add a private customer
* Add an organization customer

### DIFF
--- a/src/common/mutations/createProfile.ts
+++ b/src/common/mutations/createProfile.ts
@@ -4,7 +4,7 @@ export const CREATE_NEW_PROFILE_MUTATION = gql`
   mutation CREATE_NEW_PROFILE(
     $firstName: String!
     $lastName: String!
-    $addresses: [CreateAddressInput]
+    $addresses: [CreateAddressInput]!
     $email: String!
     $phone: String!
   ) {
--- a/src/common/mutations/createProfile.ts
+++ b/src/common/mutations/createProfile.ts
@@ -4,7 +4,7 @@ export const CREATE_NEW_PROFILE_MUTATION = gql`
   mutation CREATE_NEW_PROFILE(
     $firstName: String!
     $lastName: String!
-    $addresses: [CreateAddressInput]!
+    $addresses: [CreateAddressInput!]!
     $email: String!
     $phone: String!
   ) {
--- a/src/features/customerForm/AddCustomerFormContainer.tsx
+++ b/src/features/customerForm/AddCustomerFormContainer.tsx
@@ -15,33 +15,14 @@ const AddCustomerFormContainer = ({ onCancel, onSubmit, refetchQueries }: AddCus
   const createNewCustomer = useCreateNewCustomer(refetchQueries);
 
   const handleSubmit = (values: CustomerFormValues) => {
-    const {
-      address,
-      businessId,
-      city,
-      customerGroup,
-      email,
-      firstName,
-      lastName,
-      organizationName,
-      phone,
-      postalCode,
-    } = values;
-
-    const customerInfo = {
-      address: address,
-      businessId: businessId,
-      companyName: organizationName,
-      email: email,
-      firstName: firstName,
-      lastName: lastName,
-      municipality: city,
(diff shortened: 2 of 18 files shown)


src/common/mutations/createProfile.ts
```diff
@@ -4,7 +4,7 @@ export const CREATE_NEW_PROFILE_MUTATION = gql`
   mutation CREATE_NEW_PROFILE(
     $firstName: String!
     $lastName: String!
-    $addresses: [CreateAddressInput]
+    $addresses: [CreateAddressInput]!
```
ahie, line 7:
```suggestion
    $addresses: [CreateAddressInput!]!
```

So the generated type would resolve to

```
addresses: CreateAddressInput[]
```

instead of 

```
addresses: (CreateAddressInput | null)[];
```

my bad for originally messing this up.

TatuArvela: Sure, added a commit

src/features/customerForm/AddCustomerFormContainer.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export interface AddCustomerFormProps
+  extends Omit<FormProps<CustomerFormValues>, 'initialValues' | 'onCreate' | 'onDelete' | 'refetchQueries'> {
+  refetchQueries?: PureQueryOptions[] | string[];
+}
+
+const AddCustomerFormContainer = ({ onCancel, onSubmit, refetchQueries }: AddCustomerFormProps) => {
+  const createNewCustomer = useCreateNewCustomer(refetchQueries);
+
+  const handleSubmit = (values: CustomerFormValues) => {
```
Sartonon, line 17: this function might be easier to read like this:

```
  const handleSubmit = (values: CustomerFormValues) => {
    createNewCustomer({
      ...values,
      companyName: values.organizationName,
      municipality: values.city,
      organizationType: mapCustomerGroupAsOrganizationType(values.customerGroup),
      phoneNumber: values.phone,
      zipCode: values.postalCode,
    });
    onSubmit?.(values);
  };
```

src/features/customerList/utils.ts
```diff
@@ -75,3 +41,39 @@ function getBerthLeases(profile: ProfileNode): CustomerListBerthLeases[] | undef
...
+      if (profileEdge?.node) {
+        const profile = profileEdge.node;
+        const boats = getBoats(profile) ?? [];
+        const applications = getApplications(profile) ?? [];
+        const berthLeases = getBerthLeases(profile) ?? [];
+
+        const { organization, primaryAddress } = profile;
+
+        const profileData: CustomerData = {
+          address: organization?.address ?? primaryAddress?.address,
```
Sartonon, line 57: could organization.address be empty string? Does this work like expected then?

TatuArvela: The nullish coalescing operator still returns the left operand if it is an empty string, so this works exactly like intended: if the customer profile has an organization, the organization's address will be shown, even if it is empty.

Sartonon: Yeah, just making sure. I wanted to know if organization.address is empty string, do we want to use primaryAddress.address.

src/features/customerForm/AddCustomerFormContainer.tsx
TatuArvela: Indeed it would be 👍 The `createNewCustomer` function's `customerInfo` param will contain extra values like this, but since the function picks only the ones specified in the type, this is not a problem

Sartonon: Great :)

Sartonon (approved):

ahie (approved):
LGTM! :shipit:

PR #288, opened 2020-10-07
VEN-773, VEN-884 | Less aggressive customer form validation, unified modal behavior

TatuArvela:
## Description :sparkles:

* Make CustomerForm validate only on submit
* Make all modals close by clicking outside and remove redundant rendering conditions

### Closes :no_good_woman:

* [VEN-773](https://helsinkisolutionoffice.atlassian.net/browse/VEN-773): FE - Customer list: Add new customer 
* [VEN-884](https://helsinkisolutionoffice.atlassian.net/browse/VEN-884): Unified modal behavior

* Add / edit customer form should validate when and only when the submit button is clicked, and not when changing customer groups, for example
* All modals should close when clicking the gray backdrop outside them

### DIFF
--- a/src/features/customerView/billModal/BillModal.tsx
+++ b/src/features/customerView/billModal/BillModal.tsx
@@ -26,7 +26,7 @@ const BillModal = ({ bill, toggleModal, ...modalProps }: BillModalProps) => {
 
   const { contractPeriod } = bill;
   return (
-    <Modal isOpen={true} {...modalProps}>
+    <Modal isOpen toggleModal={() => toggleModal?.(false)} {...modalProps}>
       <Text as="h4" color="brand" className={styles.heading}>
         {t('customerView.customerBill.bill')}
       </Text>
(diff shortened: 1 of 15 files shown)


src/features/customerView/billModal/BillModal.tsx
```diff
@@ -26,7 +26,7 @@ const BillModal = ({ bill, toggleModal, ...modalProps }: BillModalProps) => {
 
   const { contractPeriod } = bill;
   return (
-    <Modal isOpen={true} {...modalProps}>
+    <Modal isOpen toggleModal={() => toggleModal?.(false)} {...modalProps}>
```
ahie, line 29: Nit: Isn't specifying `isOpen` here superfluous, as we override it with `{...modalProps}`?

```suggestion
    <Modal toggleModal={() => toggleModal?.(false)} {...modalProps}>
```

ahie (approved):
Nice changes, just a minor comment added.

src/features/customerView/billModal/BillModal.tsx
TatuArvela: `isOpen` actually was not in `{...modalProps}` as it was omitted in the type and not in the props. I decided to add the prop and make it like your suggestion for consistency. Also replaced `isOpen={true}` with `isOpen` for all Modals

ahie (approved):
LGTM 👍

PR #293, opened 2020-10-08
VEN-791 | Only show available berths on the offer page

ahie:
## Description :sparkles:

If a berth has already been assigned to some application, it should not be offered for other applications.

## Issues :bug:

[VEN-791](https://helsinkisolutionoffice.atlassian.net/browse/VEN-791)

- Go to "Venepaikkahakemukset", select an application that is in state "Odottaa käsittelyä"
- Select a harbor to choose berths from
- The offer table should not contain berths already associated with applications (check and compare with the full applications list for already assigned berths)


TatuArvela (approved):
LGTM! 👍

PR #297, opened 2020-10-12
VEN-887 | Disable create new user button if users with identical information exist

ahie:
## Description :sparkles:

Disables the create new user button in LinkApplicationToCustomer if users with identical information exist on the backend. Users with same first and last name as well as email count as identical. Additionally sets default filtering to be based on email instead of last name.

## Issues :bug:

[VEN-887](https://helsinkisolutionoffice.atlassian.net/browse/VEN-887)

### Automated tests :gear:️

- Updated relevant tests
- Added tests to `src/features/linkApplicationToCustomer/__tests__/LinkApplicationToCustomer.test.tsx`

- Select an application in state "Odottaa käsittelyä" with no connected user
- If the list of users contains a user with identical information, the create new user button should not appear
- If the list does not contain an identical user, the button should first appear but disappear after creating a new user

### DIFF
--- a/src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
+++ b/src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { useLazyQuery } from '@apollo/react-hooks';
+import { useLazyQuery, useQuery } from '@apollo/react-hooks';
 import { useDebounce } from 'use-debounce';
 
 import { getFilteredCustomersData } from './utils';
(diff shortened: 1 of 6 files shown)


src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
```diff
@@ -61,11 +61,28 @@ const LinkApplicationToCustomerContainer = ({
     variables: filteredCustomersVars,
   });
 
+  const exactlyFilteredCustomersVars: FILTERED_CUSTOMERS_VARS = {
+    first: 1,
+    [SearchBy.FIRST_NAME]: application.firstName,
+    [SearchBy.LAST_NAME]: application.lastName,
+    [SearchBy.EMAIL]: application.email,
+  };
```
TatuArvela, line 69: This will not tackle the situation where multiple customers are created with the same data, for some reason. If there were two applications from the same new customer and different handlers would handle those at the same time, two profiles would be created.

But, in practice that should never happen, at least not before we get verified signups.

TatuArvela (approved):
LGTM! 👍

src/features/linkApplicationToCustomer/LinkApplicationToCustomerContainer.tsx
ahie: Yes, to prevent this kind of race condition the check would have to be done on the server side. And as the creation actually does creation in two different backends through federation I don't see a reasonable solution to this without overhauling the architecture to properly support federated mutations.

PR #298, opened 2020-10-13
VEN-897 | Add autoDismiss option for hdsToast

TatuArvela:
## Description :sparkles:

Added an `autoDismiss` option to `hdsToast` which makes the toast dismiss automatically after 3 seconds.

### Closes :no_good_woman:

* [VEN-897](https://helsinkisolutionoffice.atlassian.net/browse/VEN-897): FE: admin application confirmation notifications should not last forever

These three actions should show a notification that dismisses automatically in 3 seconds:
* Sending invoices
* Deleting winter storage notices
* Saving an edited customer profile

### DIFF
--- a/src/common/toast/hdsToast.tsx
+++ b/src/common/toast/hdsToast.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { DismissableNotification, DismissableNotificationProps } from 'hds-react';
 import { toast } from 'react-toastify';
 import { v4 as uuidv4 } from 'uuid';
@@ -9,6 +9,7 @@ import i18n from '../../locales/i18n';
 import { HDSToastContainerId } from './HDSToastContainer';
 
 interface HDSToastArgs {
+  autoDismiss?: boolean;
   type: DismissableNotificationProps['type'];
   labelText: string;
   text: string;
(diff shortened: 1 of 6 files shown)


src/common/toast/hdsToast.tsx
```diff
@@ -41,10 +47,11 @@ const DismissableNotificationWrapper = ({
   );
 };
 
-const hdsToast = ({ type, labelText, text, toastId, translated = false }: HDSToastArgs) => {
+const hdsToast = ({ autoDismiss = false, type, labelText, text, toastId, translated = false }: HDSToastArgs) => {
```
Nizar-Rahme, line 50: Why don't we default it to `true` if this is the desired behavior?

TatuArvela: Good thought, I'll flip this around

src/common/toast/hdsToast.tsx
```diff
@@ -9,26 +9,32 @@ import i18n from '../../locales/i18n';
...
-const DismissableNotificationWrapper = ({
-  type,
-  labelText,
-  text,
-  toastId,
-  translated,
-}: DismissableNotificationWrapperProps) => {
+const NotificationWrapper = ({ autoDismiss, type, labelText, text, toastId, translated }: NotificationWrapperProps) => {
   const { t } = useTranslation();
+  const autoDismissTime = 3000;
```
Nizar-Rahme, line 27: For more clarity would it be better to take this outside the scope of the component function, as a constant:
```tsx
const AUTO_DISMISS_TIME = 3000;
```

src/common/toast/hdsToast.tsx
```diff
@@ -9,26 +9,32 @@ import i18n from '../../locales/i18n';
...
+
+  useEffect(() => {
+    if (autoDismiss) {
+      const timer = setTimeout(() => {
```
santtul, line 31: `react-toastify` seems to already support `autoClose` should we use that instead?

santtul: See https://fkhadra.github.io/react-toastify/autoClose

TatuArvela: This _sounds_ like a good idea, and I'm trying it out right now. However, `autoClose: 3000` does not actually seem to be doing anything, not even if I specify it for the toast container.

santtul: Yeah, did a quick research on this and it seems that the prop got broken on React update. No easy workaround, so let's go with this.

src/common/toast/hdsToast.tsx
santtul: Agree with @Nizar-Rahme we could move this as mentioned to the start of the file.

TatuArvela: Done.

santtul (approved):
Great work!

PR #312, opened 2020-10-26
VEN-942 | Fix the bug of refreshing paginated tables & upgrade 'react-table'

Nizar-Rahme:
## Description :sparkles:
- Upgrade `react-table`
- Prevent side effects when the initial `sortBy` matches the current state

### Closes :no_good_woman:
[**VEN-942:**](https://helsinkisolutionoffice.atlassian.net/browse/VEN-942) Refreshing paginated tables go back to the first page

### Automated tests :gear:️
- Fixed all relevant snapshots

- Go to one of the backend paginated tables e.g. Applications List
- Go to the second page
- Refresh the page
- The table should stay and render the second page as expected


src/common/table/Table.tsx
```diff
@@ -253,9 +253,12 @@ const Table = <D extends { id: string }>({
     gotoPage(0);
   }, [gotoPage, state.sortBy, state.filters, state.globalFilter]);
 
+  const initialSortBy = initialState?.sortBy;
+
   useEffect(() => {
+    if (initialSortBy?.[0]?.id === state.sortBy[0]?.id && initialSortBy?.[0]?.desc === state.sortBy[0]?.desc) return;
```
TatuArvela, line 259: This is a bit complicated, could this be expressed a bit more clearly?

src/common/table/__tests__/__snapshots__/Table.test.jsx.snap
```diff
@@ -8,7 +8,7 @@ exports[`Table renders normally 1`] = `
     <div
       class="table primary noMainHeader"
       role="table"
-      style="min-width:450px"
+      style="min-width:0px"
```
TatuArvela, line 11: This makes tables that overflow horizontally now display incorrectly

TatuArvela (changes requested):
There's now a problem with table sizing

src/common/table/__tests__/__snapshots__/Table.test.jsx.snap
Nizar-Rahme: I'll check if I can control it, I got these changes by merely upgrading the library.

Nizar-Rahme: So this was the culprit; they [changed the logic of calculating `min-width`](https://github.com/tannerlinsley/react-table/pull/2639/files) in `react-table` v `7.6.0`.

Nizar-Rahme: Therefore, we have to add `minWidth` to the columns that use `width` property. 
I added them to all the tables that have `width` already set.

src/common/table/Table.tsx
Nizar-Rahme: Done.

src/common/table/__tests__/__snapshots__/Table.test.jsx.snap
TatuArvela: Great 👍 Is `width` now needed for those columns?

TatuArvela: I tried it out, and it turns out it is. This is kind of messy, but let's investigate a better solution later.

TatuArvela (approved):
LGTM 👍 Rebased it

PR #318, opened 2020-11-04
VEN-924 | Enable filtering applications by name

ahie:
## Description :sparkles:

Enables filtering all application list tables (berth, ws, unmarked ws) by applicant full name.

### Closes :no_good_woman:

[VEN-924](https://helsinkisolutionoffice.atlassian.net/browse/VEN-924)

### Automated tests :gear:️

- `src/common/tableTools/applicationTableTools/__tests__/ApplicationTableTools.test.tsx`
- Updated relevant snapshots

- Go to any of the application pages
- A name filter should be present in the upper right corner of the table
- Typing in a name should filter the list to only show applications with matching applicant name
- It should be possible to filter by both first and last name (can also be partial string matches) at the same time and the ordering and case should not matter, e.g. "John Doe", "Doe John", "john doe" and "ohn do" should match an application that has first name "John" and last name "Doe"


src/common/tableTools/applicationTableTools/ApplicationTableTools.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { TextInput } from 'hds-react';
+
+import { ApplicationStatus } from '../../../@types/__generated__/globalTypes';
+import Select from '../../select/Select';
+import { APPLICATION_STATUS } from '../../utils/constants';
+import styles from './applicationTableTools.module.scss';
+
+interface ApplicationStateTableToolsProps {
```
TatuArvela, line 10: This needs to be renamed

TatuArvela (approved):
Great job 👍 One naming comment, but I guess you'll fix it in the next PR

src/common/tableTools/applicationTableTools/ApplicationTableTools.tsx
ahie: Yes, looking into how to combine the two new table tools recently introduced for the next PR.

PR #335, opened 2020-11-17
Update posted date in PrintAllStickersButton

ahie:


src/features/unmarkedWsNoticeList/printAllStickersButton/PrintAllStickersButton.tsx
```diff
@@ -1,18 +1,37 @@
...
   );
+  const [setStickersPosted] = useMutation<SET_STICKERS_POSTED, SET_STICKERS_POSTED_VARS>(SET_STICKERS_POSTED_MUTATION);
+
+  const onClick = () => {
+    const customers = getCustomersWithStickers(data);
+    generateAndSaveStickerPDF(customers);
+    setStickersPosted({
+      variables: {
+        input: {
+          leaseIds: customers.map((customer) => customer.leaseId as string),
```
Nizar-Rahme, line 29: What if the leaseId is `undefined`?

ahie: Those have been filtered out out `getCustomersWithStickers`. If it has a sticker number it has a lease, and if it has a lease the leaseId cannot be null per the schema.

Nizar-Rahme: Fair enough! 👍

Nizar-Rahme (approved):

PR #338, opened 2020-11-19
VEN-985 | Fix select rows in paginated tables

Nizar-Rahme:
## Description :sparkles:
- Implemented a custom hook `usePreserveSelect` for storing table selection in Recoil
- Fix non-updated state bug in `ListActions` component and remove redundant args
- Add usePreserveSelect to ApplicationList
- Add usePreserveSelect to UnmarkedWsNoticeList

### Closes :no_good_woman:
[**VEN-985**](https://helsinkisolutionoffice.atlassian.net/browse/VEN-985): FE - Fix multi-select rows on paginated tables

### Automated tests :gear:️
https://github.com/City-of-Helsinki/berth-reservations-admin/blob/1de9c1814c67612ef7649d4dd840e5682dc39282/src%2Fcommon%2Futils%2F__tests__%2FusePreserveSelect.test.tsx#L21

- Go to `Venepaikkahakemukset` view
- Select some rows from different pages in the table
- To prove that the selection is preserved you can select an action like `Lähetä tarjous` and it should perform the action on all the selected rows
- Going back and forth between pages should NOT affect the selection in `Venepaikkahakemukset` page
- The same behavior is also implemented in the `Nostojärjestysilmoitukset` view.


ahie (commented):
Excellent work! Just wondering about one thing. The preserve hook seems to have two responsibilities:

- Keep track of data for rows selected on different (table) pages. This fixes the issue that list actions can only be performed for rows on the same page.
- Keep track of selection for different (application) pages. This is an UX improvement for specific pages.

Could the first point be fixed in the table component intself? Otherwise you always need to remember to use the preserve hook with BE paginated tables, or else the selection is broken.

ahie (approved):
One open comment/discussion left, but I think we should merge as-is to get the PR queue resolved.

Nizar-Rahme:
>Could the first point be fixed in the table component itself? 
This is a great suggestion, I am on it and I'd like to get it merged in the same PR.

Nizar-Rahme:
Going to merge it as is since it turns out that moving the state to the Table component would add so much complexity. 
I can open a draft PR for the experiments of the other approach.

PR #346, opened 2020-12-04
VEN-1027 | Disable 'edit additional services' button if invoice is sent

TatuArvela:
## Description :sparkles:

* Disable 'edit additional services' button if invoice is sent

### Closes :no_good_woman:

* [VEN-1027](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1027): Invoice 'edit additional services' button should be inactive if invoice is sent

### DIFF
--- a/src/features/applicationView/berthOfferCard/BerthOfferCard.tsx
+++ b/src/features/applicationView/berthOfferCard/BerthOfferCard.tsx
@@ -84,6 +84,7 @@ const BerthOfferCard = ({
       className={className}
       customerEmail={customerEmail}
       invoicingDisabled={!berthInvoicingFeatureFlag()}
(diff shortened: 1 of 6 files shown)


santtul (changes requested):
Seems to work, but should we indicate that the button is disabled in the UI somehow (e.g. by graying it) for better usability?

TatuArvela:
> Seems to work, but should we indicate that the button is disabled in the UI somehow (e.g. by graying it) for better usability?

I replaced the button with an actual small button, which is grayed when it is disabled

Nizar-Rahme (approved):
LGTM!

ahie (approved):
LGTM 👍

PR #348, opened 2020-12-07
Fix browser tests on GitHub CI

TatuArvela:
## Description :sparkles:

* Added a simple .exists() check which makes the browser test wait a bit before clicking

### DIFF
--- a/browser-tests/navigateAndView.ts
+++ b/browser-tests/navigateAndView.ts
@@ -17,6 +17,8 @@ test('Navigate and view basic data', async (t) => {
   await t
     .click(navigation.applicationsExpandable)
     .click(navigation.applications)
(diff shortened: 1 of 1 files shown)


TatuArvela:
I added a simple `.exists()` check which makes the test wait a it before clicking. It seems to fix the problem, but is there a better solution?

santtul (approved):
👍

Nizar-Rahme:
Would [wait method](https://devexpress.github.io/testcafe/documentation/reference/test-api/testcontroller/wait.html) be any good in this scenario?

santtul:
Usually the click should do the wait under the hood, weird that it's not working in this case.

santtul:
@Nizar-Rahme wait(s) with defined timeouts should be avoided.

PR #349, opened 2020-12-07
Remove running browser tests on Windows

TatuArvela:
## Description :sparkles:

Solution for browser test issues. Since `windows-latest` is the only one with difficult timing issues in browser tests, let's just remove it.


santtul (approved):
👍

TatuArvela:
Closed because GitHub waits on a removed check which can't be ignored

PR #350, opened 2020-12-08
VEN-952 | Recurring Invoices

Nizar-Rahme:
## Description :sparkles:
- Implement the recurring invoices page and its subcomponents wrapped in a container 
- Implement a common `DataSummary` component

- Click on `Kausilaskut` in the sidebar
- The summary section shows how many customers will be invoiced 
- To invoice the customers, click on the `Lähetä venepaikkojen laskut` button which opens a modal
- From the modal select the due date and hit `Lähetä`

## Screenshots :camera_flash:
image


src/features/recurringInvoices/RecurringInvoices.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </Section>
+      <p>
+        <Trans
+          i18nKey="recurringInvoices.failureInstructions"
+          values={{ failedInvoicesCount: failedInvoicesCount ?? '-' }}
+        >
+          ... <Text as="strong">failedInvoicesCount</Text> ...
+        </Trans>
+      </p>
+      <Table
```
ahie, line 80: This table might need backend pagination for performance reasons in production, but that can be done in a separate PR later.

src/features/recurringInvoices/RecurringInvoices.tsx
```diff
@@ -0,0 +1,99 @@
...
+const RecurringInvoices = ({
+  loading,
+  dataSummary,
+  failedInvoicesCount,
+  failedInvoicesData,
+  handleSend,
+}: RecurringInvoicesProps) => {
+  const { t } = useTranslation();
+
+  const failedInvoicesCols: Column<FailedInvoices>[] = [
```
ahie, line 40: We might actually want a separate column for the date when the failure occurred. But that should be a separate task.

src/features/customerView/CustomerViewContainer.tsx
```diff
@@ -40,7 +40,10 @@ const CustomerViewContainer = () => {
   const [openInvoice, setOpenInvoice] = useState<Invoice>();
   const { t } = useTranslation();
   const { id } = useParams<{ id: string }>();
-  const { loading, data } = useQuery<INDIVIDUAL_CUSTOMER>(INDIVIDUAL_CUSTOMER_QUERY, { variables: { id } });
+  const { loading, data } = useQuery<INDIVIDUAL_CUSTOMER>(INDIVIDUAL_CUSTOMER_QUERY, {
+    variables: { id },
+    fetchPolicy: 'no-cache',
```
ahie, line 45: Was this query being fetched from the cache previously? I thought we already configured apollo to not use cache? I think there was a PR that fixed some issue by removing lines like this previously.

src/common/utils/auth.ts
```diff
@@ -0,0 +1,7 @@
+import authService from '../../features/auth/authService';
+
+export const getProfileToken = () => {
```
ahie, line 3: Optional: could this be simply moved to the `AuthService` itself?

ahie (approved):
Excellent work! Just added a few comments that can be addressed in separate PRs if needed. I'd say :shipit:.

src/features/recurringInvoices/RecurringInvoices.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </Section>
+      <Section>
+        <DataSummary labelValuePairs={dataSummary} />
+      </Section>
+      <p>
+        <Trans
+          i18nKey="recurringInvoices.failureInstructions"
+          values={{ failedInvoicesCount: failedInvoicesCount ?? '-' }}
+        >
+          ... <Text as="strong">failedInvoicesCount</Text> ...
```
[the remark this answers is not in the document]
Nizar-Rahme: Exactly, this is just for making the count bold.

src/features/customerView/CustomerViewContainer.tsx
Nizar-Rahme: This is interesting, It is configured to use `network-only`, but for some reason, Apollo Client was acting weird; when I go to a single customer page, the `useQuery` hook would `return` undefined for the data, even though the actual data is fetched correctly (by observing the requests).
Therefore, this line was needed here to work around the issue.

src/features/recurringInvoices/RecurringInvoices.tsx
Nizar-Rahme: That makes sense, but there is no such field exposed from the backend, is there?

src/features/recurringInvoices/RecurringInvoices.tsx
Nizar-Rahme: 👍

src/locales/fi.json
```diff
@@ -598,6 +599,28 @@
...
+    "sendInvoices": "Lähetä venepaikkojen laskut",
+    "failedInvoicesTable": {
+      "header": "Kausilasku ei lähetetty",
+      "name": "Nimi",
+      "harbor": "Satama",
+      "failureReason": "Virheen syy",
+      "emptyState": "Ei laskuja"
+    },
+    "summary": {
+      "customers": "Asiakkaita",
```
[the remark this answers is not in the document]
Nizar-Rahme: Done.

src/common/utils/auth.ts
Nizar-Rahme: Yup, I will move it in a follow-up PR.

PR #351, opened 2020-12-08
VEN-952 | Add missing env variables for gitlab pipeline

Nizar-Rahme:
## Description :sparkles:
- Update GitLab and Docker config files to reflect the new changes of env variables

### Related :handshake:
[VEN-952](https://helsinkisolutionoffice.atlassian.net/browse/VEN-952): FE: Add recurring invoicing page


ahie (approved):

ahie:
Looks like the browser tests are missing these variables also

Nizar-Rahme:
I am wondering why! Browser tests should read them from .env file, they passed in the previous PR 🤔

Nizar-Rahme:
Even [this step](https://gitlab.com/City-of-Helsinki/KuVa/github-mirrors/berth-reservations-admin/-/jobs/896906284) is failing on GitLab, any insights?

ahie:
That just looks like a random failure to me:
> Error from server (BadRequest): container "anders-deploy-app" in pod "qa-r-fix-missing-m82ea6-review-6bf7d94dc4-d876m" is waiting to start: trying and failing to pull image

ahie:
Are GHA browser tests using `TEST_ENV_URL` or `LOCAL_ENV_URL`? Looks like they're using `TEST_ENV_URL` for which the login is broken before this fix.

ahie:
So the fix would be to change `getEnvUrl` in `browser-tests/utils/settings.ts` to always return localhost. Otherwise PRs that add new browsertests for new features in the same PR will also be broken. I think the `TEST_ENV_URL` is an artifact from before GHA when travis only ran them as a cron job.

Nizar-Rahme:
Let's try that! 👍

Nizar-Rahme:
But in this case, don't we need to run a local instance in order to run the browser tests?

ahie:
Yes that would also need to be done

ahie:
But I don't see any other option to getting the browser tests to work correctly for PRs.

PR #356, opened 2020-12-09
Run browser tests against local instances

Nizar-Rahme:
## Description :sparkles:
- Run browser tests in Github Actions against local instances


package.json
```diff
@@ -70,7 +70,7 @@
     "pre-push": "yarn types:generate && yarn lint && yarn stylelint",
     "browser-test": "testcafe \"chrome --window-size='1920,1080'\" browser-tests/",
     "browser-test:local": "TEST_ENV=local testcafe \"chrome --window-size='1920,1080'\" browser-tests/",
```
TatuArvela, line 72: After these changes these seem to work identically. There is no longer a way to run browser tests against the staging environment, but is there a valid use case for that anyway?

browser-tests/utils/settings.ts
```diff
@@ -1,7 +1,6 @@
 // @ts-ignore
 require('dotenv').config({ path: '.env.development.local' });
 
-const TEST_ENV_URL = 'https://venepaikka-admin.test.kuva.hel.ninja';
 const LOCAL_ENV_URL = 'http://localhost:3000';
```
TatuArvela, line 4: Maybe we should allow setting this in `.env.development.local`?

TatuArvela (approved):
I wrote a couple of comments about reduced functionality, but looks good. The delay is maybe a bit long, but eh 🤷

browser-tests/utils/settings.ts
Nizar-Rahme: Done!

package.json
Nizar-Rahme: Yeah, couldn't come up with any valid use case for running the tests against the staging environment.

Dockerfile
```diff
@@ -68,6 +68,7 @@ ARG REACT_APP_TUNNISTAMO_SCOPE_BERTHS
 ARG REACT_APP_TUNNISTAMO_SCOPE_PROFILE
 ARG REACT_APP_TUNNISTAMO_URI
 ARG REACT_APP_BERTH_API_URL
+ARG REACT_APP_LOCAL_ENV_URL
```
TatuArvela, line 71: You need to add `ENV REACT_APP_LOCAL_ENV_URL $REACT_APP_LOCAL_ENV_URL` under this.
This could also be higher in this ARG list near `REACT_APP_ENV`

Nizar-Rahme: Browser tests are configured to read its environment variables from `.env.development.local`, therefore, I moved this variable there, and add it to GH Secrets.

TatuArvela:
It seems login does not work

ahie:
> It seems login does not work

Tunnistamo only allows requests from certain sources, e.g. `localhost:3000`, staging and prod are whitelisted

ahie (changes requested):
The server needs to be started separately before calling `yarn browser-test:ci`

Nizar-Rahme:
@ahie `--app \"yarn start\"` option in the `yarn browser-test:ci` script runs the server.

Nizar-Rahme:
>It seems login does not work
@TatuArvela It's because when you added `BROWSER_TESTS_LOCAL_ENV_URL` to `.env` file, it overrides the secrets from GitHub, therefore, I removed it from `.env`. 
If you want to run the tests locally, you should add a `.env.development.local` file (it's listed already in `.gitignore`) which contains the missing variables, namely:
```
BROWSER_TESTS_UID=
BROWSER_TESTS_PWD=
BROWSER_TESTS_LOCAL_ENV_URL=
```

TatuArvela:
> > It seems login does not work
> 
> @TatuArvela It's because when you added `BROWSER_TESTS_LOCAL_ENV_URL` to `.env` file, it overrides the secrets from GitHub, therefore, I removed it from `.env`.
> If you want to run the tests locally, you should add a `.env.development.local` file (it's listed already in `.gitignore`) which contains the missing variables, namely:
> 
> ```
> BROWSER_TESTS_UID=
> BROWSER_TESTS_PWD=
> BROWSER_TESTS_LOCAL_ENV_URL=
> ```

Interesting, good to have that resolved. I'm aware of `.env.development.local`, it just did not seem like it'd override that way as the GH config defines the values from secrets and the script uses `.env.development.local` for the config. [dotenv](https://github.com/motdotla/dotenv) recommends using just one `.gitignore`d `.env` file, which we could investigate, but let's not get into that now.

ahie (approved):
LGTM! The `--app-init-delay` might be too short, but we can tweak that later if it turns out to cause issues.

PR #357, opened 2020-12-10
VEN-1026 | Add pier price category display and input

TatuArvela:
## Description :sparkles:

* Add pier price category display and input

### Closes :no_good_woman:

* [VEN-1026](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1026): FE - Set pricing tier for pier

### DIFF
--- a/src/features/harborView/forms/pierForm/PierForm.tsx
+++ b/src/features/harborView/forms/pierForm/PierForm.tsx
@@ -14,6 +14,9 @@ import FormHeader from '../../../../common/formHeader/FormHeader';
 import ConfirmationModal from '../../../../common/confirmationModal/ConfirmationModal';
 import Checkbox from '../../../../common/checkbox/Checkbox';
 import Button from '../../../../common/button/Button';
+import RadioButton from '../../../../common/radioButton/RadioButton';
+import { PriceTier } from '../../../../@types/__generated__/globalTypes';
+import { priceTierToString } from '../../utils';
 
 interface PierFormProps extends FormProps<Pier> {
   suitableBoatTypeOptions: BoatType[];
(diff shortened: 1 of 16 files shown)


src/features/harborView/utils.ts
```diff
@@ -125,3 +127,16 @@ export const getMaps = (data: INDIVIDUAL_HARBOR | undefined): Map[] => {
...
+
+export const priceTierToString = (priceTier: PriceTier | null): string => {
+  switch (priceTier) {
+    case PriceTier.TIER_1:
+      return '1';
+    case PriceTier.TIER_2:
+      return '2';
+    case PriceTier.TIER_3:
+      return '3';
+    default:
```
ahie, line 139: Optional: could remove the default here so if anything changes it will be caught as an error during build.

src/features/harborView/forms/pierForm/PierForm.tsx
```diff
@@ -127,6 +132,26 @@ const PierForm = ({
             </div>
           </Grid>
 
+          <div className={styles.priceCategories}>
```
ahie, line 135: Nit: mixed terminology (price tiers / price categories)

ahie (commented):
LGTM so far. Just two optional comments added.

src/features/harborView/utils.ts
TatuArvela: Changed. `PierProperties` is now rendered with `pier.priceTier !== null ? priceTierToString(pier.priceTier) : '-'` because `priceTier` is nullable in the API

src/features/harborView/forms/pierForm/PierForm.tsx
TatuArvela: Good catch, changed

ahie (approved):
LGTM!

PR #366, opened 2020-12-16
Increase server init delay time

Nizar-Rahme:
## Description :sparkles:
Increate the server init delay to 90 seconds

## Issues :bug:
Failing browser tests

Only the first browser test keeps failing, so I am suspecting that the issue has to do with the server not being ready.


ahie (approved):
👍

PR #375, opened 2021-01-07
Optimize query used in HarborView

ahie:
This patch breaks up the nested query used in HarborView into separate queries, decreasing the number of joins performed on the backend. Requires the following BE PR to be merged: https://github.com/City-of-Helsinki/berth-reservations/pull/434.


Nizar-Rahme (approved):
LGTM!

PR #376, opened 2021-01-08
Scrub access_token from breadcrumb data sent to sentry

ahie:


src/index.tsx
```diff
@@ -14,6 +14,12 @@ const { REACT_APP_SENTRY_DSN, REACT_APP_SENTRY_ENVIRONMENT } = process.env;
 Sentry.init({
   dsn: REACT_APP_SENTRY_DSN,
   environment: REACT_APP_SENTRY_ENVIRONMENT,
+  beforeBreadcrumb(breadcrumb) {
+    if (JSON.stringify(breadcrumb.data).includes('access_token')) {
+      delete breadcrumb.data;
```
Nizar-Rahme, line 19: Wondering if we should omit the whole `data` property or getting rid of only `access_token` would do the trick.

TatuArvela: Any update on this? Should we omit more fields or merge this as it is?

TatuArvela (approved):
Let's get this merged already 👍

Nizar-Rahme:
👍

PR #401, opened 2021-03-17
Combine browser-tests with production, review and staging workflows

TatuArvela:
## Description :sparkles:

* Combine browser-tests with production, review and staging workflows
* This should reduce random browser test failures considerably


Nizar-Rahme (approved):
LGTM! 🥇

PR #410, opened 2021-04-16
VEN-1257 | Make offer action permissions more granular

Nizar-Rahme:
## Description :sparkles:
- Hide invoice action based on the order status

### Related :handshake:
[VEN-1257](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1257)

### DIFF
--- a/src/features/invoiceCard/InvoiceCardContainer.tsx
+++ b/src/features/invoiceCard/InvoiceCardContainer.tsx
@@ -48,17 +48,19 @@ const InvoiceCardContainer = ({
     setSelectedInvoiceAction(null);
   };
 
-  const actionsDisabled = !(order?.status === OrderStatus.ERROR || order?.status === OrderStatus.OFFERED);
-
   return (
     <>
       <InvoiceActions
-        disabled={actionsDisabled}
         selectedAction={selectedInvoiceAction}
         actions={[
           {
             value: 0,
             label: t('invoiceCard.markAsPaid.label'),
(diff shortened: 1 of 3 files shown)


src/features/invoiceCard/InvoiceCardContainer.tsx
```diff
@@ -48,17 +48,19 @@ const InvoiceCardContainer = ({
...
-        disabled={actionsDisabled}
         selectedAction={selectedInvoiceAction}
         actions={[
           {
             value: 0,
             label: t('invoiceCard.markAsPaid.label'),
+            disabled: !(
+              order?.status === OrderStatus.ERROR ||
+              order?.status === OrderStatus.OFFERED ||
+              order?.status === OrderStatus.DRAFTED
```
TatuArvela, line 62: Is there a reason that this is duplicated in `invoicingDisabled`? This could probably be extracted

TatuArvela (approved):
One suggestion, LGTM 👍

src/features/invoiceCard/InvoiceCardContainer.tsx
Nizar-Rahme: "Mark as paid" would change the status to be paid manually while the "invoicing disabled" is for sending an offer, aka it changes the status to be "offered" and these conditions happen to be the same for both but that won't be necessarily the case, therefore, I didn't extract them to one variable.

PR #429, opened 2021-05-07
Increment package version to 2.0.0

TatuArvela:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "berth-reservations-admin",
-  "version": "0.1.0",
+  "version": "2.0.0",
```
Nizar-Rahme, line 3: That's a big leap 😁

Nizar-Rahme (approved):

PR #434, opened 2021-05-10
Remove Storybook

TatuArvela:
## Description :sparkles:

We are not using our stories or maintaining them, so let's remove them.


Nizar-Rahme (approved):
Worth mentioning, iirc it was because of a conflict with the Storybook we had to give `_colours.sass` partial a British name 😁

PR #449, opened 2021-11-17
VEN-1367 | Admin can give a marked WS place to customer without an application

Jokinen:
## Description :sparkles:

Adds an userflow that allows an admin to add a winter storage lease for a customer without an application.

Mimics the approach taken with berths and introduces slight refactors in order to make code sharing easier.

Also

* Fixes issue with WinterStorageLanguage which looked unrelated to these changes

Flow

First: Access customer page

Expect to be able to add a winter storage lease without an application.

I checked the happy path and that the cancellation paths work.

## Screenshots :camera_flash:

Screenshot 2021-11-17 at 11 25 06
Screenshot 2021-11-17 at 11 26 01
Screenshot 2021-11-17 at 11 26 13
Screenshot 2021-11-17 at 11 26 36

* I think that it'd make more sense to select the boat first and then use its data to mark unsuitable areas in the dropdown. With one test user it was difficult to find a suitable boat. I didn't do this as the design did not ask for it.
* Lots of boats lack names in which case the boat select lists empty labels--imo an "Unnamed" label would be better than an empty selection--I wonder if we could pull other fields from the boat to give at least a somewhat descriptive value. But again, I felt that this was out of scope.


src/features/customerView/createLease/queries.ts
```diff
@@ -24,3 +24,28 @@ export const CREATE_LEASE_HARBORS_QUERY = gql`
     }
   }
 `;
+
+export const CREATE_WINTER_STORAGE_LEASE_MODAL_QUERY = gql`
+  query CREATE_WINTER_STORAGE_LEASE_MODAL($customerId: ID!) {
+    profile(id: $customerId, serviceType: BERTH) {
```
Nizar-Rahme, line 30:
```suggestion
    profile(id: $customerId, serviceType: BERTH) {
      id
```
Sometimes Apollo cache starts acting up when the response doesn't have the id

src/features/customerView/createLease/CreateWinterStorageLeaseModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+function useWinterStorageAreas({ customerId }: UseWinterStorageAreasConfig) {
+  return useQuery<CREATE_WINTER_STORAGE_LEASE_MODAL, CREATE_WINTER_STORAGE_LEASE_MODAL_VARS>(
+    CREATE_WINTER_STORAGE_LEASE_MODAL_QUERY,
+    {
+      fetchPolicy: 'no-cache',
+      variables: { customerId },
+    }
+  );
+}
```
Nizar-Rahme, line 99: Optional, you may put this query in a separate module.

src/features/applicationView/utils.ts
```diff
@@ -39,6 +44,21 @@ export const mapBerthApplicationLanguageToLanguage = (
...
+): Language | null => {
+  switch (applicationLanguage) {
+    case 'FI':
+      return Language.FINNISH;
+    case 'SV':
+      return Language.SWEDISH;
+    case 'EN':
+      return Language.ENGLISH;
+    default:
+      return null;
```
Nizar-Rahme, line 58: Why don't we fall back to Finnish for example?

Nizar-Rahme (approved):
A couple of cosmetic comments, otherwise, it looks good to me 👍

src/features/applicationView/utils.ts
Jokinen: The function this helper approximates returned `null` and there's a testcase for that behaviour.

As I only made this change to fix a type issue I didn't want to introduce any behavioural changes.

In that sense, although the lines are new, this code isn't new. I only added support for the WinterStorageLanguage enum into it.

PR #454, opened 2022-02-03
Require the phone number in Create customer -form

nikomakela:
VEN-1466


nikomakela:
Don't know why the types were updated so much (2nd commit). Because there are not so many unit tests for them, it was hard to catch all the issues raised because of updating them. The pre-push hook didn't allow me to push my changes before updating the schema.

Jokinen (approved):

PR #455, opened 2022-02-04
Fix the Apollo issue of the create berth switch offer

nikomakela:
VEN-1454. Apollo threw an error when the 2nd level node of individual application query did not contain an id. An cache issue which is described here: https://www.apollographql.com/docs/react/caching/cache-configuration/#default-identifiers


nikomakela:
There is an alternative or complementary fix #456

Jokinen (approved):

PR #456, opened 2022-02-04
Fix the Apollo issue of the create berth switch offer

nikomakela:
VEN-1454. An cache issue which is described here: https://www.apollographql.com/docs/react/caching/cache-configuration/#default-identifiers

There is an alternative for this fix https://github.com/City-of-Helsinki/berth-reservations-admin/pull/455


nikomakela:
#455 selected as a better alternative, because it does not disable the cache.

PR #457, opened 2022-02-10
VEN-1437 | As an admin I can filter query set with a filter form

Jokinen:
## Description :sparkles:

Adds filters to customer list which allows users to be filtered based on various conditions.

After filters have been selected, they should be shown in a list that can be controlled.

Integration with backend may yet require some changes, we'll see.

### Closes :no_good_woman:

[VEN-1437](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1437)

### Automated tests :gear:️

Contains a relatively heavy tests that check the happy path of filling out inputs.

## Screenshots :camera_flash:

Screenshot 2022-02-16 at 12 56 56
Screenshot 2022-02-16 at 12 57 03


src/features/customerList/__tests__/CustomerListContainer.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const mocks = config?.mocks ?? [];
+
+  render(<CustomerListContainer />, {
+    mocks,
+  });
+
+  return {
+    toggleFilters: () => {
+      userEvent.click(screen.getByRole('button', { name: 'Rajaa' }));
+    },
```
Sartonon, line 21: These helper selector could be global in this module. Then there is no need to destructure return value from `setup()`. Just like we use the global screen object in all instead of destructuring return value from render for selectors.

But this is not very important, not need to cahnge this necessarily :)

src/features/customerList/__tests__/CustomerListContainer.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+  ].forEach(({ label, type }) => {
+    switch (type) {
+      case 'multiselect':
+        expect(screen.getByRole('combobox', { name: label })).toBeInTheDocument();
+        break;
+      case 'select':
+        expect(screen.getByRole('combobox', { name: label })).toBeInTheDocument();
+        break;
+      case 'checkbox':
+        expect(screen.getByRole('checkbox', { name: label })).toBeInTheDocument();
```
Sartonon, line 118: `screen.queryByRole` is more suitable for `.toBeInDocument` assertion. But also `getByRole` will tell you which combobox wasn't found. In a way (code execution) `.toBeDocument` is redundant when using `getByRole`.

In this situation `getByRole` might be better.

src/features/customerList/customerListTableFilters/CustomerListTableFiltersForm.tsx
```diff
@@ -0,0 +1,426 @@
...
+      'harborIds',
+      'pierId',
+      'berthId',
+      'winterStorageGridAreaIds',
+      'winterStoragePlaceId',
+      'winterStorageAreaIds',
+      'startDate',
+      'endDate',
+      'moreThanOneBerthOrWinterStorage'
+    );
```
Sartonon, line 137: Maybe something like `onResetAll` would make it clearer that this resets the whole form.

src/features/customerList/customerListTableFilters/CustomerListTableFiltersForm.tsx
```diff
@@ -0,0 +1,426 @@
...
+
+    onChange(fakeChangeEvent);
+  };
+
+  return (
+    <Combobox<Option>
+      {...rest}
+      multiselect={true}
+      disabled={disabled || !options}
+      placeholder={!disabled && !options ? t('customerList.message.loadingOptions') : placeholder}
```
Sartonon, line 352: Can be confusing as to why `!disabled && !options` means that options are loading. isLoading prop or something like that would make the loading logic little bit more understandable.

Sartonon: Now reader of the code has to read this component to understand when input is in loading state

src/features/customerList/customerListTableFilters/CustomerListTableFiltersFormContainer.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+const CustomerListTableFiltersContainer = ({ onFormClose }: Props) => {
+  const { t } = useTranslation();
+  const [listTableFilters, setListTableFilters] = useListTableFilters();
+  const [filters, setFilters] = useRecoilState(filtersAtomFamily(listTableFilters));
+  const queryByHarborId = filters.harborIds.length === 1 ? filters.harborIds[0] : null;
+  const queryByWinterStorageGridAreaId =
+    filters.winterStorageGridAreaIds.length === 1 ? filters.winterStorageGridAreaIds[0] : null;
+  const { data } = useQuery<FILTER_OPTIONS>(FILTER_OPTIONS_QUERY, {
+    fetchPolicy: 'no-cache',
```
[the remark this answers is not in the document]
Sartonon: Tested with `fetchPolicy: 'cache-first'` and I think it worked well. No need to wait for sub filters to load every time.

src/features/customerList/customerListTableFilters/CustomerListTableFiltersList.tsx
```diff
@@ -0,0 +1,267 @@
...
+    moreThanOneBerthOrWinterStorage: (_: FilterEntryValue) => '',
+  };
+
+  const removeFilter = (entry: FilterEntry) => {
+    const nextFilters = getEntriesAsFilters(
+      filtersAsEntries.filter(([key, value]) => !(key === entry[0] && value === entry[1]))
+    );
+
+    setFilters(nextFilters);
+  };
```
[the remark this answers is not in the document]
Sartonon: You can come up with even better variables names :D

src/features/customerList/customerListTableFilters/CustomerListTableFiltersList.tsx
```diff
@@ -0,0 +1,267 @@
...
+            </li>
+          );
+        })}
+      </ul>
+      <button
+        type="button"
+        className={styles.resetFiltersButton}
+        onClick={() => {
+          clearFilters();
+        }}
```
Sartonon, line 176: Could be:
```suggestion
        onClick={clearFilters}
```

Sartonon (commented):
Overall looking veeery good 👍 

Added some improvements ideas and small comments.

src/features/customerList/customerListTableFilters/CustomerListTableFiltersFormContainer.tsx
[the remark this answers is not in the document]
Jokinen: I couldn't determine why `no-cache` was preferred in this application so I mimicked the approach. I don't mind dropping it here.

src/features/customerList/customerListTableFilters/CustomerListTableFiltersList.tsx
```diff
@@ -0,0 +1,267 @@
...
+        if (startDate && !endDate) {
+          return `${format(startDate, 'yyyy')}-`;
+        }
+
+        if (!startDate && endDate) {
+          return `-${format(endDate, 'yyyy')}`;
+        }
+
+        if (startDate && endDate) {
+          return `${format(startDate, 'yyyy')} - ${format(endDate, 'yyyy')}`;
```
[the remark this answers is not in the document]
Jokinen: Yes. Screenshot from design:

Screenshot 2022-02-16 at 8 06 59

Not 100% that this `yyyy` format wasn't left in by accident, because the format in which the date is given by the user changed from a year into a date at some point.

src/features/customerList/__tests__/CustomerListContainer.test.tsx
Jokinen: I think I know what you mean. I had this same thought for some time, but I've maybe changed my mind since.

The expectation will be truthy 99% of the time the codes executes that far because most of the times that something is wrong, the getter will throw.

As I see it, the expectation (and `toBeInTheDocument`) is there more for the human. It helps communicate that we are making an assertion on this line. Without `expect` the line could seem bare to some eyes. The error `getBy` yields is helpful in pointing the dev towards the correct direction when they have to adjust tests.

Checking out the readme for `jest-dom`, their first example of this helper uses a `getBy` selector:
https://github.com/testing-library/jest-dom#tobeinthedocument

Also, testing library itself has this approach documented:
https://testing-library.com/docs/guide-disappearance/#2-using-waitfor

It gives `queryBy` as an option when asserting that the element isn't in the document (which is the most common usecase for `queryBy`):
https://testing-library.com/docs/guide-disappearance/#nottobeinthedocument

I look at the `getBy` selectors' throwing as a way to make fixing tests easier for the developer. That's why I don't mind using them in this way anymore.

src/features/customerList/customerListTableFilters/CustomerListTableFiltersFormContainer.tsx
```diff
@@ -0,0 +1,211 @@
...
+}
+
+const CustomerListTableFiltersContainer = ({ onFormClose }: Props) => {
+  const { t } = useTranslation();
+  const [listTableFilters, setListTableFilters] = useListTableFilters();
+  const [filters, setFilters] = useRecoilState(filtersAtomFamily(listTableFilters));
+  const queryByHarborId = filters.harborIds.length === 1 ? filters.harborIds[0] : null;
+  const queryByWinterStorageGridAreaId =
+    filters.winterStorageGridAreaIds.length === 1 ? filters.winterStorageGridAreaIds[0] : null;
+  const { data } = useQuery<FILTER_OPTIONS>(FILTER_OPTIONS_QUERY, {
```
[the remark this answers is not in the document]
Jokinen: I now find the options from a hook that I have declared within this component file. I think that it removes your pain while not requiring a lot of other changes.

Personally I would guess that a nice overall fix here could be to have for instance `<HarbourMultiSelect />` or `<PierSelect harbourId={...} />` but that might take too much time to refactor at this point. I attempted to conform to the container/component pattern in a very literal sense because the codebase isn't that well know to me--but perhaps these kinds of component would be OK.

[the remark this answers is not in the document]
Sartonon: In my opinion container component can also be a child component and I think you agree on that. 

But I think that refactoring this to your suggestion at this point might take too much time but I also think that current solution is acceptable 👍  :)

src/features/customerList/__tests__/CustomerListContainer.test.tsx
Sartonon: I completely agree and was thinking the same when commenting (which is why my comment was unnecessary :D). `getByRole` for debuggability and `.toBeInTheDocument` for readability.

src/features/auth/authService.ts
```diff
@@ -68,6 +68,19 @@ class AuthService {
...
+
+    if (tokens) {
+      const tokensObject = JSON.parse(tokens);
+      const { REACT_APP_TUNNISTAMO_SCOPE_PROFILE: profileScope } = process.env;
+
+      return profileScope ? tokensObject[profileScope] : null;
+    } else {
+      return null;
+    }
+  }
```
Sartonon, line 82: auth.ts file already has this helper. Maybe one or the other should be deleted?

src/features/customerList/customerListTableFilters/labels/CustomerListTableFiltersListWinterStorageAreaLabel.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const CustomerListTableFiltersListHarborKeyword = ({ id }: Props) => {
+  const { data } = useQuery<FILTER_WINTER_STORAGE_AREA_LABEL, FILTER_WINTER_STORAGE_AREA_LABEL_VAR>(
+    FILTER_WINTER_STORAGE_AREA_LABEL_QUERY,
+    {
+      variables: {
+        winterStorageAreaId: id,
+      },
+    }
+  );
```
Sartonon, line 23: Good place to have cache-first policy

src/features/customerList/customerListTableFilters/labels/CustomerListTableFiltersListHarborLabel.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface Props {
+  id: string;
+}
+
+const CustomerListTableFiltersListHarborKeyword = ({ id }: Props) => {
+  const { data } = useQuery<FILTER_HARBOR_LABEL, FILTER_HARBOR_LABEL_VAR>(FILTER_HARBOR_LABEL_QUERY, {
+    variables: {
+      harborId: id,
+    },
+  });
```
Sartonon, line 20: Good place to have cache-first policy

src/features/customerList/customerListTableFilters/CustomerListTableFilters.tsx
```diff
@@ -0,0 +1,35 @@
...
+          className={styles.expandButton}
+          onClick={() => {
+            setFiltersExpanded((prev) => !prev);
+          }}
+        >
+          {t('customerList.message.filter')} <IconAngleDown aria-hidden="true" />
+        </button>
+      </div>
+      {areFiltersExpanded && <CustomerListTableFiltersFormContainer onFormClose={() => setFiltersExpanded(false)} />}
+      {!areFiltersExpanded && someFilterIsActive && <CustomerListTableFiltersList filters={listTableFilters} />}
```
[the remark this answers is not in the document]
Jokinen: I didn't chain ternaries here, but I used another approach to make it easier to perceive the branching.

src/features/customerList/customerListTableFilters/CustomerListTableFiltersList.tsx
```diff
@@ -0,0 +1,186 @@
...
+  });
+  const winterStoragePlaceQuery = useQuery<FILTER_WINTER_STORAGE_PLACE_LABEL, FILTER_WINTER_STORAGE_PLACE_LABEL_VAR>(
+    FILTER_WINTER_STORAGE_PLACE_LABEL_QUERY,
+    {
+      skip: !filters.winterStoragePlaceId,
+      variables: {
+        winterStoragePlaceId: filters.winterStoragePlaceId ?? '',
+      },
+    }
+  );
```
Sartonon, line 67: These could still use cache first fetch policy to avoid repeated requestt when opening and closing filters

src/features/customerList/customerListTableFilters/CustomerListTableFilters.tsx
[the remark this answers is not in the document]
Sartonon: 👍 fine for me

Sartonon (approved):
Amazing job 🥳 🏅

Jokinen:
Rebased and fixed commit history.

I'll add commits from feature/ven-1437-proposal into this branch now.

PR #459, opened 2022-02-11
Ignore profile queries' does not exist errors from a toaster

nikomakela:
VEN-1468 VEN-1470. Save profile-query's OBJECT_DOES_NOT_EXIST_ERROR from a toaster. The missing profiles are already informed other ways and it does not make sense to pop up an error notification multiple times a page load.

image

Names are already formatted:
image


Sartonon (approved):
Looks good.

PR #466, opened 2022-02-17
VEN-1443 | Add table export functionality to customers and application views

Sartonon:
## Description :sparkles:

Adds table export functionality.

Looping the results might not be needed but lets see how the API evolves.

### DIFF
--- a/src/common/utils/__tests__/useTableExport.test.ts
+++ b/src/common/utils/__tests__/useTableExport.test.ts
@@ -5,7 +5,6 @@ import * as Apollo from '@apollo/react-hooks';
 import { useTableExport } from '../useTableExport';
 import { exportClient } from '../../../app/exportClient';
 import * as downloadFile from '../downloadFile';
-// import { getBerthToken } from '../auth';
 
 jest.mock('../auth');
 
--- a/src/common/utils/__tests__/useTableExport.test.ts
+++ b/src/common/utils/__tests__/useTableExport.test.ts
@@ -1,0 +1,92 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { renderHook } from '@testing-library/react-hooks';
+import * as Apollo from '@apollo/react-hooks';
+
+import { useTableExport } from '../useTableExport';
+import { exportClient } from '../../../app/exportClient';
+import * as downloadFile from '../downloadFile';
+// import { getBerthToken } from '../auth';
+
+jest.mock('../auth');
+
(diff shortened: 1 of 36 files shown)


src/common/utils/__tests__/useTableExport.test.ts
```diff
@@ -0,0 +1,92 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { renderHook } from '@testing-library/react-hooks';
+import * as Apollo from '@apollo/react-hooks';
+
+import { useTableExport } from '../useTableExport';
+import { exportClient } from '../../../app/exportClient';
+import * as downloadFile from '../downloadFile';
+// import { getBerthToken } from '../auth';
```
Jokinen, line 8: FYI

src/features/applicationView/ApplicationViewContainer.tsx
```diff
@@ -87,12 +88,14 @@ const ApplicationViewContainer = () => {
...
     });
-  const handleLinkCustomer = (customerId: string) =>
+  const handleLinkCustomer = (customerId: string) => {
     linkCustomer({
       variables: {
         input: { id, customerId },
       },
-    });
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+    }).catch(() => {});
```
Jokinen, line 97: Intentional?

Jokinen (approved):
The technical approach looks really really good here 👍

I'm mainly wondering about the UX. Seems like the label may be PoC ("Vie..").

But I'm also thinking about the "Toiminnot" dropdown 🤔 It's open by default which makes it a bit of a non issue, but it's not the most descriptive label. Of course, this is an admin UI so the admin can be taught so not a huge problem.

I'm also thinking whether there should be some kind of indicator that the excel is being downloaded/built?

Sartonon:
> The technical approach looks really really good here 👍
> 
> I'm mainly wondering about the UX. Seems like the label may be PoC ("Vie..").
> 
> But I'm also thinking about the "Toiminnot" dropdown 🤔 It's open by default which makes it a bit of a non issue, but it's not the most descriptive label. Of course, this is an admin UI so the admin can be taught so not a huge problem.
> 
> I'm also thinking whether there should be some kind of indicator that the excel is being downloaded/built?

Now that I was testing all the export endpoints, I noticed missing loading indicator too. There should be some indicator for that. "Vie..." is from the designs but not the best label...

PR #468, opened 2022-02-21
VEN-1431 | Fix pagination in customer list page

Jokinen:
## Description :sparkles:

Fixes issues with pagination:

* Changing page reset filters
* Removing all filters doesn't clear persisted filters from session storage

Queries on customers page can be forced through by typing for instance `lind` in the surname filter box.


nikomakela (approved):
Looks good to me. Tested some of the filters and they were reset when page was changed.

_However, the filters seems to not work yet https://helsinkisolutionoffice.atlassian.net/browse/VEN-1477 and the sorting does not work either https://helsinkisolutionoffice.atlassian.net/browse/VEN-1476. We could collaborate with those. Haven't had time to test them earlier._

PR #470, opened 2022-02-22
Fix customer xlsx export 

charn:
API renamed `profile_token` field to `profileToken` to be consistent with the GraphQL side.

Refs VEN-1443


nikomakela (approved):
consistency is good! So is the change.

Jokinen (approved):
Looks good 👍

PR #471, opened 2022-02-22
VEN-1443 | Fix customer export filtering

Jokinen:
## Description :sparkles:

Hooks up new filters with the customer exporter.

Now only users matching the filters should be included in the generated excel.

## Issues :bug:

[VEN-1443](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1443)


nikomakela (approved):
👍

PR #474, opened 2022-02-28
VEN-1479 | As an admin I should select a harbor first in the customers listing to avoid huge forbidden query to Open-City-Profile

Jokinen:
## Description :sparkles:

- To use new filters, user must select harbour
- If no new queries are used, the pre-existing `profiles` query is used
   - Users can be searched by their last names as before
- Notifies user about limited search

⚡️ If the user searches by ticket number or registration number, they must select a harbour. This is bad UX. Is it bad enough that we need to address it with this temporary adjustment?

## Issues :bug:
[VEN-1479](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1479)

## Screenshots :camera_flash:

Screenshot 2022-02-28 at 15 34 11
Screenshot 2022-02-28 at 15 34 18
Screenshot 2022-02-28 at 15 34 59


nikomakela:
I could not export the results in excel. Should it have been possible? It works in https://venepaikka-admin.test.kuva.hel.ninja/
image

nikomakela (approved):
Good! Minor thing, but the export button could also be disabled in impossible cases

PR #479, opened 2022-04-26
VEN-1501: Adds previous lease to the admin UI

melniiv:
## Description :sparkles:
Adds previous lease customer name to the Customer column to the Harbors table view
If there is an active lease -> show current customer name as link
If there is no active lease but previous lease exists -> show previous customer name as greyed out text
Otherwise cell is empty

## Screenshots :camera_flash:
image


nikomakela (approved):
Didn't have time to test it, but the code changes looks good.

Jokinen (approved):

PR #481, opened 2022-04-28
VEN-1340: Winter storage area edit

melniiv:
## Description :sparkles:
Adds the edit form and edit button to the individual winter storage area

## Screenshots :camera_flash:
image

### DIFF
--- a/src/@types/__generated__/globalTypes.ts
+++ b/src/@types/__generated__/globalTypes.ts
@@ -766,6 +766,27 @@ export interface UpdateWinterStorageApplicationInput {
   clientMutationId?: string | null;
 }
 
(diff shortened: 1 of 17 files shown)


nikomakela (commented):
I don't think the WWW-URL should be a mandatory field. Other than that OK! Code changes looks good.

nikomakela:
> I don't think the WWW-URL should be a mandatory field. Other than that OK! Code changes looks good.

Should it also be visible some where. Cannot find it from the UI.

melniiv:
its not for some reason in the card, i was also wondering why, but we can add it later if there is a need

nikomakela (approved):
approved, but the url should not be required!

PR #490, opened 2022-11-10
VEN-921 Search By Name

kristianjokela:
## Description :sparkles:
Currently you can only search customers by either `firstName` or `lastName`. This combines both names to a single `name`, that's only used internally, and splits the `name` back to `firstName` or `lastName` for the query in the backend. This allows "free" name queries like "Me Ma" for "Matti Meikäläinen", etc.. Only limitation is that the names have to be inputted in the form of lastName firstName but there's a placeholder in the input field indicating this.

### Closes :no_good_woman:
VEN-921

## Screenshots :camera_flash:
Screenshot 2022-11-10 at 10 32 40

### DIFF
--- a/src/features/customerList/CustomerListContainer.tsx
+++ b/src/features/customerList/CustomerListContainer.tsx
@@ -25,6 +25,7 @@ import { InvoicingType } from '../../@types/__generated__/globalTypes';
 export enum SearchBy {
   FIRST_NAME = 'firstName',
   LAST_NAME = 'lastName',
+  NAME = 'name',
   EMAIL = 'email',
   ADDRESS = 'address',
   STICKER_NUMBER = 'stickerNumber',
@@ -34,7 +35,7 @@ export enum SearchBy {
 
 const searchByAtom = atom<SearchBy>({
   key: 'CustomerListContainer_searchByAtom',
-  default: SearchBy.LAST_NAME,
+  default: SearchBy.NAME,
 });
 
 const searchValAtom = atom<string>({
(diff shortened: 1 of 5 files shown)


src/features/customerList/CustomerListContainer.tsx
```diff
@@ -65,6 +64,20 @@ const orderBySelector = selector<string | undefined>({
   },
 });
 
+// Backend supports queries for names in the form of firstName, lastName only so
+// this transforms SearchBy.NAME to firstName, lastName.
+const transformSearchByName = (searchBy: SearchBy, value: string) => {
```
nikomakela, line 69: Works in most of the cases, but what when the name is in 3 parts (or more)? Should the lenght be checked first? If the lenght is 3 or more, which parts belongs to first and which to the lastname?

nikomakela (approved):
approved, if the multipart name is a risk that you want to take

src/features/customerList/CustomerListContainer.tsx
kristianjokela: Now it's pretty stupid, but I was thinking since the first input is always the last name, it might be ok for now to just let admins type enough of the last name to get close enough to meaningful results (since results update while typing), and then use first name to "refine" the search. Now anything after second part is just ignored. 

So essentially now the problems would come from two part (without hyphen) last names - IF the admins are fast enough to type both parts - , and last names with some kind of prefix like "Af".. Maybe the small prefixes might be good idea to include with first part though - that could cause issues actually.

kristianjokela: We discussed this with the team and came to the conclusion that this is good enough for now for most common use cases, but we'll keep the last name and first name options for more exotic names (added on the next commit: https://github.com/City-of-Helsinki/berth-reservations-admin/pull/490/commits/35db2b64fa9fccaba6ad1cc9ad20f51651b1b2a5)

PR #492, opened 2022-11-16
VEN-1534 Search companies by name

kristianjokela:
## Description :sparkles:
This allows admins to search companies directly by company name. Currently you can only search companies by the contact person, which is not optimal if you don't know who the contact person is.

This implementation will simply fetch all the organizations and filter them on the client side. So far it works fast (ie. not slower than any other query) on dev with (as time of writing) 339 results. However, if there's a sudden surge of companies to the platform, say thousands, we might need to implement the search on the backend.

Also, since the company search uses `customerGroup=company` to get all the companies, there's a collision with `customerGroup` filter; ie. user could select search by company name and then add a `customerGroup=private`. Now this is solved by resetting `searchBy` to `name` and displaying a message to the user what just happened.

### Closes :no_good_woman:
VEN-1534

## Screenshots :camera_flash:
Screenshot 2022-11-16 at 10 50 33


nikomakela (approved):
approved in current situation, but would it make more sense to add a custom resolver in the gateway? Something that could try to make several subqueries and combine the results?

https://www.apollographql.com/docs/apollo-server/data/resolvers/

kristianjokela:
> approved in current situation, but would it make more sense to add a custom resolver in the gateway? Something that could try to make several subqueries and combine the results?
> 
> https://www.apollographql.com/docs/apollo-server/data/resolvers/
> 
> Probably also the cache could be handled better there, like the organisations could be prefetched in cache.

Yeah, this is the initial version, but definitely the idea is to get this done properly on the gateway. Especially the overlap with the filter makes this a bit less optimal than initially thought - so I'm not completely happy about this solution either.

PR #493, opened 2022-11-23
Ven-1537 Update customer filter

kristianjokela:
## Description :sparkles:
Seems the customer filter doesn't timeout as much as before. Queries are slow, but nevertheless filtering does seem to work without `harbourId` limitation too (in test). So this removes the limitation, but informs users about the slowness of the queries. Whenever the queries start working faster, the notification can be removed (and other related tickets finally resolved).

Also, there was a bug how the `invoicingTypes` was handled on the query, which most of the time resulted on empty results, this also fixes that.

### Closes :no_good_woman:
VEN-1537

### Related :handshake:
https://helsinkisolutionoffice.atlassian.net/browse/VEN-1480
https://helsinkisolutionoffice.atlassian.net/browse/HP-1317
https://helsinkisolutionoffice.atlassian.net/browse/VEN-1481

### DIFF
--- a/src/features/customerList/CustomerListContainer.tsx
+++ b/src/features/customerList/CustomerListContainer.tsx
@@ -68,18 +68,28 @@ const orderBySelector = selector<string | undefined>({
   },
 });
 
-// Backend supports queries for names in the form of firstName, lastName only so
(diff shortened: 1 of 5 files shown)


nikomakela (approved):
Looks good to me

PR #494, opened 2022-11-28
VEN-983 Search by sticker number and season

kristianjokela:
## Description :sparkles:
This allows filtering searches by both sticker number and season combination. Backend with `stickerSeason` needs to be deployed first since this relies on the new filter.

Since queries to berthProfile are slow, triggering queries from input box too fast seems to cause errors from the backend (the responses will come eventually, but users see error notifications), there's slower debounce rate on input for queries to `berthProfile`. Users will see that UI is loading from the first key press so users won't notice the difference (since the query is slow anyway...).

Also, we need to discuss if the separate stickernumber -only filter is needed, for now the task was to keep both.

### Closes :no_good_woman:
VEN-983
### Related :handshake:
https://helsinkisolutionoffice.atlassian.net/browse/VEN-1538

## Screenshots :camera_flash:
Screenshot 2022-12-02 at 9 55 30

### DIFF
--- a/src/features/customerList/CustomerListContainer.tsx
+++ b/src/features/customerList/CustomerListContainer.tsx
@@ -31,6 +31,7 @@ export enum SearchBy {
   EMAIL = 'email',
   ADDRESS = 'address',
   STICKER_NUMBER = 'stickerNumber',
+  STICKER_NUMBER_SEASON = 'stickerNumberSeason',
   BOAT_REGISTRATION_NUMBER = 'boatRegistrationNumber',
   INVOICING_TYPE = 'invoicingType',
 }
@@ -85,6 +86,18 @@ const transformSearchBy = (searchBy: SearchBy, value: string): SearchByTransform
   if (searchBy === SearchBy.INVOICING_TYPE) {
     return {
       invoicingTypes: [value],
+    };
+  }
+  if (searchBy === SearchBy.STICKER_NUMBER_SEASON) {
+    if (value.indexOf(' ') === -1) {
+      return {
+        stickerNumber: value,
+      };
+    }
+    const [stickerNumber, season] = value.split(' ');
+    return {
+      stickerNumber,
+      stickerSeason: season,
     };
   }
   return {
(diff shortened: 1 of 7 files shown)


src/features/customerList/CustomerListContainer.tsx
```diff
@@ -87,6 +88,18 @@ const transformSearchBy = (searchBy: SearchBy, value: string): SearchByTransform
       invoicingTypes: [value],
     };
   }
+  if (searchBy === SearchBy.STICKER_NUMBER_SEASON) {
+    if (value.indexOf(' ') === -1) {
```
nikomakela, line 92: trivial: I know this is the old and common way to check the (non-)presense, but I would still use `includes` instead, because it says it more clear what it does. Most probably it does the very same thing, since both are pure vanilla JS.

```suggestion
    if (!value.includes(' ')) {
```

src/features/customerList/CustomerListContainer.tsx
```diff
@@ -108,7 +121,15 @@ const CustomerListContainer = () => {
...
   const [displayNotification, setDisplayNotification] = useState(false);
 
-  const [debouncedSearchVal] = useDebounce(searchVal, 500, {
+  const debounceWait = [
+    SearchBy.STICKER_NUMBER,
+    SearchBy.STICKER_NUMBER_SEASON,
+    SearchBy.BOAT_REGISTRATION_NUMBER,
+  ].includes(searchBy)
+    ? 1200
+    : 500;
```
nikomakela, line 130: What are these? They could be named and set in a constant variable. Trivial thing anyway.

src/features/customerList/tableTools/CustomerListTableTools.tsx
```diff
@@ -102,7 +107,7 @@ const CustomerListTableTools = <T extends string>({
               className={styles.searchInput}
               id="searchSimilarCustomers"
               value={searchVal}
-              placeholder={searchBy === 'name' ? `${t('common.lastName')} ${t('common.firstName')}` : undefined}
+              placeholder={selectedSearchByOption?.placeholder}
```
nikomakela, line 110: good

src/locales/fi.json
```diff
@@ -320,7 +320,9 @@
       "phone": "Puhelinnumero",
       "email": "Sähköpostiosoite",
       "comment": "Huomiot",
-      "selectInvoicingTypePaperInvoice": "Laskutus paperisena"
+      "selectInvoicingTypePaperInvoice": "Laskutus paperisena",
+      "stickerNumberAndSeason": "Tarra ja kausi",
+      "stickerNumberAndSeasonPlaceholder": "Numero vuosi/vuosi"
```
nikomakela, line 325: vuosi/vuosi? 🤔

nikomakela (approved):
Small notes, but approved. GJ!

src/features/customerList/CustomerListContainer.tsx
kristianjokela: I'm not sure what to call them :D  - They're the default 500ms and extended 1200ms for slower queries; I guess I can add them as constants. Ideally we can remove them when profile queries get faster.

src/locales/fi.json
kristianjokela: This might need some discussion still:) Seasons are in the form "2020/2021" So it's "start year / end year", but it should fit in the input box nicely.

kristianjokela: OK, now they're "Nro alkuvuosi/loppuvuosi" which is more clear. Seems the input box doesn't _need_ top be 25% width, so the box will stretch to fit whole string now.

PR #502, opened 2022-12-16
Pipeline YML files

Vitals9367:


mikkovihonen (approved):
This pull request is too big to review properly. Approving as we need to get going with the deployment, but adding a ticket to review with the team.

PR #509, opened 2023-02-28
Enable more caching

kristianjokela:
Since there are still some issues with customer list queries, enable caching on apollo client to speed up the UI. This won't fix the underlying issue that the combined queries from our backend and Helsinki profile are just slow, but at least it will make the UI feel faster and for repeated queries, there's less waiting.


terovirtanen:
**Venepaikat-admin-ui branch is deployed to platta: https://venepaikat-admin-ui-pr509.dev.hel.ninja** :rocket::rocket::rocket:

PR #512, opened 2023-07-05
Upgrade the Node version to 14

nikomakela:
VEN-1553

- Update the node types package
- Update the docker base image

PR #513, opened 2023-07-05
Upgrade the Node version to 16

nikomakela:
VEN-1553

- Upgrade the minimum amount of the dependencies so that they support Node 16
- Update the docker base image

NOTE: The node v. 14 upgrade is in #512 . This branch is based on that.


terovirtanen (approved):
Node version should update on devops pipeline variable too

nikomakela:
This change should be tested carefully. Some dependencies might not work properly even if they don't complain about anything.

nikomakela:
There are more changes needed than we have time to for now. The Node 14 changes are merged now in https://github.com/City-of-Helsinki/berth-reservations-admin/pull/512

PR #555, opened 2024-08-20
VEN-1575 | Make updateable

karisal-anders:
### fix: "yarn types:generate" & .env URLs, update generated types

The `.env` file had obsolete URLs in it, copied the currently used
values from the test environment into it.

Type generation (i.e. "yarn types:generate") failed because of:
```
Error: Cannot use GraphQLInputObjectType "ApproveOrderMutationInput"
from another module or realm.

Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of
other relied on modules, use "resolutions" to ensure only one version is
installed.

https://yarnpkg.com/en/docs/selective-version-resolutions

Duplicate "graphql" modules cannot be used at the same time since
different versions may have different capabilities and behavior. The
data from one version used in the function from another could produce
confusing and spurious results.
```

To make only a single version of graphql package to be used, update the
@types/apollo-upload-client, apollo-upload-client and graphql packages
to compatible versions.

Now "yarn types:generate" should work out-of-the-box after copying
`.env` to `.env.development.local` as per the README instructions.

Also updated the generated types resulting in only comment changes.

refs VEN-1575 (Need to be able to update the generated types)

### chore: run "yarn prettier", "yarn lint --fix" and "yarn stylelint --fix"

refs VEN-1575

## Issues :bug:

[VEN-1575](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1575)

[VEN-1575]: https://helsinkisolutionoffice.atlassian.net/browse/VEN-1575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


karisal-anders:
Replaced by PR #556

